PR #19, opened 2022-10-11
Add a service open feature flag

felixclack:
### Context

We want to be able to control access to the service using a feautre
flag.

We have a new library, govuk_feature_flags, that implements the actual
feature flag function. This change introduces that as a dependency and
configures it for use here.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

Screenshot 2022-10-13 at 10 11 17 am


tvararu (approved):
:boom:

PR #28, opened 2022-10-13
Fix stylesheet naming collision

felixclack:
### Context

`application.css` gets overwritten by the design system dependency when the
assets get compiled. The result of this is that app defined styles never
get delivered to the client.

The easiest solution is to use a different name for the stylesheet to avoid
this happening.

`cssbundling-rails` will enhance the `rails spec` command provided by
`rspec-rails` to ensure assets are compiled correctly in the test
environment. Switching to using the rake tasks means we can write less
setup code to run tests.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


tvararu (approved):
This is good, one to remember to upstream to the template if it all works correctly :+1:

PR #32, opened 2022-10-14
Add a page with links to the relevant referral form

felixclack:
### Context

When someone completes the eligibility screener, they should see
guidance on the appropriate referral form they need to use.

Currently, the forms are hosted on GOV.UK so we opted to hotlink to them
and accept the maintenance that comes with hotlinking. This is
preferable over having to host them in the service and keep them in sync
with the GOV.UK ones.

This change introduces a simple session based method to persist teh
eligibility check across multiple steps of the form.

It is anticipated that this will evolve into something like the
solutions on Find and Access.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

Screenshot 2022-10-14 at 12 09 08 pm
Screenshot 2022-10-14 at 12 08 26 pm

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -1,4 +1,8 @@
 class PagesController < ApplicationController
+  def complete
+    @eligibility_check = EligibilityCheck.find(session[:eligibility_check_id])
+  end
+
   def start
   end
 end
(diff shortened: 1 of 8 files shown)


tvararu (approved):
:boom:

PR #42, opened 2022-10-17
Add the serious misconduct screener question

felixclack:
One of the questions in the screener asks if you are reporting serious
misconduct.

Choosing "no" prompts you to report the issue somewhere else.

Both "yes" and "I'm not sure" take you to the next step in the flow.

The "I'm not sure" option also displays a hint as to what to do next.

This question was created using the recently added form generator.

There is no back link present, nor any object orchestrating the question
order. These will be added in subsequent PRs.

I was using this question as a qay of verifying the form generator was
configured appropriately for the refer project.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

| Name | Screen |
| - | - |
| Default | Screenshot 2022-10-17 at 12 42 07 pm |
| Error | Screenshot 2022-10-17 at 12 42 46 pm |
| Not sure |  Screenshot 2022-10-17 at 12 43 10 pm |
| Choosing No | Screenshot 2022-10-17 at 12 42 57 pm |

### DIFF
--- a/app/controllers/reporting_as_controller.rb
+++ b/app/controllers/reporting_as_controller.rb
@@ -9,7 +9,7 @@ class ReportingAsController < ApplicationController
       ReportingAsForm.new(reporting_as_params.merge(eligibility_check:))
     if @reporting_as_form.save
       session[:eligibility_check_id] = eligibility_check.id
-      redirect_to you_should_know_path
+      redirect_to serious_path
     else
       render :new
     end
--- a/app/controllers/serious_misconduct_controller.rb
+++ b/app/controllers/serious_misconduct_controller.rb
@@ -1,0 +1,28 @@
+class SeriousMisconductController < ApplicationController
+  def new
+    @serious_misconduct_form = SeriousMisconductForm.new
+  end
+
+  def create
+    @serious_misconduct_form =
+      SeriousMisconductForm.new(
+        serious_misconduct_form_params.merge(eligibility_check:)
+      )
+    if @serious_misconduct_form.save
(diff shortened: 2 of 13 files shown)


app/controllers/serious_misconducts_controller.rb
```diff
@@ -0,0 +1,28 @@
+class SeriousMisconductsController < ApplicationController
```
tvararu, line 1: Should this be singular?

```suggestion
class SeriousMisconductController < ApplicationController
```

spec/system/service_wizard_spec.rb
```diff
@@ -30,6 +34,25 @@
...
+    when_i_press_continue
+    then_i_see_a_validation_error
+
+    when_i_choose_not_sure
+    then_i_see_the_not_sure_hint
+
+    when_i_choose_no
+    when_i_press_continue
+    then_i_see_the_not_serious_misconduct_page
+  end
```
tvararu, line 54: This pattern is against a rule in our testing styleguide, we should try to have system specs that only have one scenario: https://github.com/DFE-Digital/refer-serious-misconduct/blob/main/docs/testing_styleguide.md#rules

Our test suite in Find grew pretty unwieldy because we didn't apply this rule earlier... let's try it differently this time.

Perhaps for now, we can simply extend the main scenario with these assertions, and as it grows, we can consider extracting out separate scenarios/files for testing things like validations or dead-ends.

tvararu (approved):

PR #46, opened 2022-10-18
Add a support interface for eligibility checks

felixclack:
### Context

We want a page in the support interface to provide some visibility into
the eligibility checks that are created by this service.

Taking inspiration from [Find](https://github.com/DFE-Digital/find-a-lost-trn/pull/53), this change introduces
a basic implementation of a list of `EligibilityCheck`s.

The list is limited to the most recent 100 records. When the list
starts to exceed that limit in production, we can look at adding
pagination.

Also at this point, we haven't added any data anonymisation to this page
as there is nothing personal or specific stored yet. When that changes
we can revisit this decision.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

Screenshot 2022-10-18 at 10 24 56 am

### DIFF
--- a/app/views/support_interface/eligibility_checks/index.html.erb
+++ b/app/views/support_interface/eligibility_checks/index.html.erb
@@ -1,0 +1,33 @@
+<% content_for :page_title, 'Eligibility Checks' %>
+
+<h1 class="govuk-heading-l">
+  Eligibility Checks
+  <span class="govuk-caption-l">
+    Ordered by last updated, showing <%= "#{[100, EligibilityCheck.count].min} of #{EligibilityCheck.count}" %> total
+  </span>
+</h1>
+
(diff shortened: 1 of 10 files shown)


app/views/support_interface/eligibility_checks/index.html.erb
```diff
@@ -0,0 +1,33 @@
...
+    end
+
+    summary_list.row do |row|
+      row.key(text: 'Updated at')
+      row.value(text: eligibility_check.updated_at)
+    end
+
+    summary_list.row do |row|
+      row.key(text: 'Reporting as')
+      row.value(text: eligibility_check.reporting_as)
```
tvararu, line 25: Can we `.titleize` this and the serious misconduct answer later?

tvararu (approved):

PR #47, opened 2022-10-18
Add the teaching in England question

felixclack:
The screener includes a question to determine if the person being
referred was teaching in England at the time of alleged offence.

This helps establish jurisdiction.

If the answer is no, then we display a no jurisdiction message and end
the screener at that point.

If the answer is yes or not sure, then we continue with the next
question.

With the addition of this question, the need for an object to
orchestrate the order of the questions is becoming more pressing.

This will happen in an upcoming PR and again will take inspiration from
the implementation in Find.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

Screenshot 2022-10-18 at 11 39 39 am


tvararu (approved):
:+1:

PR #56, opened 2022-10-19
Add back links

felixclack:
### Context

Each question we present in the screener should give people a way to go
back to the previous one.

For now, the back links are simple and can always return to the
referring page.

### Checklist

- [ ] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


tvararu (approved):
Yep, this will need iteration, later :smile:

PR #64, opened 2022-10-21
Add the "is a teacher" question

felixclack:
### Context

We want to determine if the allegation is about a teacher.

This question introduces a new fork and exit page for the flow.

While doing this the complexity of the enforce question order class was
growing because it was handling the non-answerable pages.

To simplify, I moved the logic for these exit pages to the controller
that precedes them. This allows the question ordering class to return to
a single responsibility.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

Screenshot 2022-10-21 at 10 33 56 am

### DIFF
--- a/app/controllers/unsupervised_teaching_controller.rb
+++ b/app/controllers/unsupervised_teaching_controller.rb
@@ -9,7 +9,11 @@ class UnsupervisedTeachingController < ApplicationController
         unsupervised_teaching_form_params.merge(eligibility_check:)
       )
     if @unsupervised_teaching_form.save
-      next_question
+      if eligibility_check.unsupervised_teaching?
+        next_question
+      else
+        redirect_to(no_jurisdiction_unsupervised_path)
+      end
     else
       render :new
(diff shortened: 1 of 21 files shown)


app/controllers/unsupervised_teaching_controller.rb
```diff
@@ -9,7 +9,11 @@ def create
         unsupervised_teaching_form_params.merge(eligibility_check:)
       )
     if @unsupervised_teaching_form.save
-      next_question
+      if !eligibility_check.unsupervised_teaching?
```
steventux, line 12: Could this condition be inverted

```ruby
if eligibility_check.unsupervised_teaching?
  next_question
else
  redirect_to(no_jurisdiction_unsupervised_path)
end
```
to avoid an _if not_ statement? Or use `unless`?
Not a blocker.

steventux (approved):
:+1: :rocket:

PR #69, opened 2022-10-25
Add a question about an informal complaint

felixclack:
If answering the screener as a member of the public then you should see
a question about whether you have made an informal complaint already.

If not, we provide information about how to do so and still present the
option to continue with the screener.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

Screenshot 2022-10-25 at 9 42 06 am

Screenshot 2022-10-25 at 9 42 14 am


steventux (approved):
:+1: LGTM (failing spec seems to be an unrelated timeout error?)

tvararu (approved):

PR #72, opened 2022-10-25
Basic Devise setup for Staff users

malcolmbaig:
### Context

We want to authenticate Staff users in a way similar to how we've done so in Find a lost TRN and Apply for QTS.

https://trello.com/c/igamP4FJ

Extract / copy implementation from Find. 

This PR sets up the staff model, controllers and views.

Subsequent PRs will handle:

- Invites
- A mangement interface for staff users
- An alternative, basic auth strategy for accessing the support interface.

### Guidance to review

- It looks like a lot, but all of this is code is either directly lifted (via cherry-pick) or produced by a generator.
- Per-commit review highly recommended. 
- No specs or any properly testable functionality in this PR, so a once-over of the code should suffice.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


tvararu (approved):
:ship:

PR #81, opened 2022-10-26
Add sidekiq

malcolmbaig:
### Context

As per ADR #2, we use Sidekiq and Redis for out of band jobs. The specific requirement here is to send Devise emails via Notify, but this setup will also be used for other jobs further down the line.

Add the sidekiq gem and all required config to use these tools in dev and test. A subsequent PR will ensure Redis is available in production.

### Guidance to review

No real testing required, this will be exercised once Devise invites are built.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,6 +20,7 @@ gem "pg", "~> 1.4"
 gem "propshaft"
 gem "puma", "~> 5.0"
 gem "rails", "~> 7.0.3"
+gem "sidekiq"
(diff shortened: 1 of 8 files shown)


tvararu (approved):
One additional thing you could do is set up a critical and low priority queue:

https://github.com/DFE-Digital/find-a-lost-trn/blob/main/config/sidekiq.yml#L9-L12

PR #86, opened 2022-10-26
Add staff management page to support

malcolmbaig:
### Context

We need somewhere we can invite staff users to access support.

- Add the staff management page to support
- Add a sign out link in the support header to facilitate the sign in / sign out flow for staff users

### Guidance to review

Can be tested in a review app or locally by navigating to /support. The Staff page is linked in the header. You'll need either zero Staff records in the database or the `staff_http_basic_auth` feature flag on to access support with the basic auth credentials.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/staff/invitations_controller.rb
+++ b/app/controllers/staff/invitations_controller.rb
@@ -38,7 +38,11 @@ class Staff::InvitationsController < Devise::InvitationsController
     end
   end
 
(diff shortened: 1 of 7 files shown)


steventux (approved):
:rocket:

PR #94, opened 2022-10-27
Referrals employer form - Personal details / their name page

steventux:
### Context

When making a referral, the details of the person being referred need to be specified.

https://teacher-misconduct.herokuapp.com/report/teacher/name

https://trello.com/c/QKgbFLPr/864-referrals-employer-form-their-name-page

image

- Namespace referrals routes
- Adds `Referrals::PersonalDetailsForm` as the basis for backing all pages in the relevant section form flow
- Adds appropriate actions, views, errors

### Guidance to review

I've gone with 1 controller + form object per section, so DOB, NiN and QTS forms will use these too.
Error messages are written in `devspeak`, possibly need content review.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


config/routes.rb
```diff
@@ -41,6 +41,15 @@
   get "/referrals/(:id)/delete", to: "referrals#delete", as: "delete_referral"
   get "/referrals/deleted", to: "referrals#deleted", as: "deleted_referral"
 
+  namespace :referrals do
```
vassyz, line 44: I think you can move the two rows above into this namespace.

app/forms/referrals/personal_details_form.rb
```diff
@@ -0,0 +1,43 @@
...
+              on: :names,
+              if: -> { name_has_changed == "yes" }
+
+    def initialize(referral:, step: :names, names: {})
+      @referral = referral
+      @step = step
+      initialize_names(names:) if step == :names
+    end
+
+    def initialize_names(names:)
```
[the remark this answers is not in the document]
steventux: If we are replaying previous answers persisted on the model it is yes, the edit template uses `form_with @personal_details_form` as the backing object.

config/routes.rb
steventux: I attempted this but it causes namespacing issues with the controller specified in the `to:` - it looks for `::Referrals::ReferralsController#delete` for example. One solution would be to move all Referrals related controllers into that module and satisfy the problem.

spec/system/referrals/user_adds_personal_details_spec.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+RSpec.feature "User adds personal details to a referral" do
```
malcolmbaig, line 4: Proper minor, but could be:

```suggestion
RSpec.feature "Personal details" do
```

app/controllers/referrals/personal_details_controller.rb
```diff
@@ -0,0 +1,31 @@
+module Referrals
+  class PersonalDetailsController < ReferralsController
```
malcolmbaig, line 2: Along the lines of something that I think worked decently in Apply for Teacher Training, we could have:

```suggestion
  class PersonalDetails::NameController < ReferralsController
```

ie—a controller and form for each step of the flow. Allows you to stick to REST-ful action names (just edit/update in this case) and simpler form logic. WDYT?

app/forms/referrals/personal_details_form.rb
```diff
@@ -0,0 +1,43 @@
+module Referrals
+  class PersonalDetailsForm
```
malcolmbaig, line 2: Linking this to the comment above, re: controller/form granularity https://github.com/DFE-Digital/refer-serious-misconduct/pull/94/files#r1009255456

config/routes.rb
steventux: @vassyz having discussed the more atomic approach of a controller per form I think you're right about tidying up these routes, something like 

```ruby
  resources :referrals, except: %i[index show] do                                                                                                                                          
      get "/delete", to: "referrals#delete", on: :member                                                                                                                                     
      get "/deleted", to: "referrals#deleted", on: :collection                                                                                                                               
                                                                                                                                                                                             
      member do                                                                                                                                                                              
        namespace :personal_details do                                                                                                                                                       
          get "/name", to: "name#edit"                                                                                                                                                       
          put "/name", to: "names#update"                                                                                                                                                    
        end                                                                                                                                                                                  
      end                                                                                                                                                                                    
    end            
 ```
 
 appears to produce the necessary routes in a bit of a tidier way, will amend.

File: app/controllers/referrals/personal_details_controller.rb (hunk 4 above)
Comment (steventux) [reply app/controllers/referrals/personal_details_controller.rb:2]: Yes agree - as discussed amending to a controller per step, much cleaner.

File: app/forms/referrals/personal_details_form.rb (hunk 2 above)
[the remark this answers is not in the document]
Comment (steventux) [reply app/forms/referrals/personal_details_form.rb]: @vassyz have amended to remove this method. Tried to be more consistent with the contact-details/email PR.

File: config/routes.rb (hunk 1 above)
Comment (steventux) [reply config/routes.rb:45]: Having discussed this I've amended here https://github.com/DFE-Digital/refer-serious-misconduct/pull/94/commits/58eca0f92f87bc0e0d885c5e736b517f388b6656

File: app/controllers/referrals/personal_details/name_controller.rb [hunk 6]
```diff
@@ -0,0 +1,39 @@
+module Referrals
+  module PersonalDetails
+    class NameController < ReferralsController
```
vassyz, line 3: I know `Style/ClassAndModuleChildren` is disabled from `rubocop-govuk`. But do we prefer a particular style? I can see the compact format(`::`) used in quite a few places. cc @malcolmbaig @tvararu

tvararu: I like the compact one better, we can override our own Rubocop yml to suggest it. :+1:

vassyz: I looked a bit into it because I was wondering why Rubocop recommends it and I think it makes sense to go with the longer format that @steventux uses in this PR. Defining nested modules can cause issues when the parent module has not already been defined. I updated my PR to have the same format as I think we can run into issues.

app/controllers/referrals/personal_details/name_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+        @personal_details_name_form =
+          NameForm.new(
+            referral:,
+            first_name: referral.first_name,
+            last_name: referral.last_name,
+            name_has_changed: referral.previous_name.present?,
+            previous_name: referral.previous_name
+          )
+
+        render "referrals/personal_details/name"
```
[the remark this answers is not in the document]
steventux: Yeah I prefer `edit` for sure, missed this in the earlier refactor - will amend.

malcolmbaig (approved):

vassyz (approved):

PR #97, opened 2022-10-28
Referrals employer form - Contact details / their email page

vassyz:
### Context

When making a referral, the email of the person being referred needs to be specified.

https://teacher-misconduct.herokuapp.com/report/teacher-contact-details/email

https://trello.com/c/fF5VwXw5/861-contact-form-add-email-address-page

Add the following page, accessible from the employer form hub:
Screenshot 2022-10-28 at 15 06 44

* Update the routes
* Generate a migration for adding the email to the referrals table.
* Add the form, including validation 
* Link the form to the referral hub

### Guidance to review

To make merging easier, I have updated this PR to be more in-line with @steventux's #94 as I initially had a slightly different approach. One thing I didn't update was to create one from model for the "Contact details" journey. It seems like Steve 
created a [PersonalDetailsForm](https://github.com/DFE-Digital/refer-serious-misconduct/pull/94/files#diff-14e5071aa6cd1cb2cfaad4dc826c6400223159eaff193c2b0182d4d8f4eace21R2). I'd rather have a [model](https://github.com/DFE-Digital/refer-serious-misconduct/pull/97/files#diff-f24b6365d33f4c1eff916a217b38b5a75c2077b24eee898f64f0cfc293f4de12R2) for every form in the "Contact details" section.

### Checklist

- [X] Attach to Trello card
- [X] Rebased main
- [X] Cleaned commit history
- [X] Tested by running locally

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -36,6 +36,8 @@ ActiveRecord::Schema[7.0].define(version: 2022_11_01_131509) do
   create_table "referrals", force: :cascade do |t|
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.boolean "email_known"
+    t.string "email_address", limit: 256
     t.string "first_name"
     t.string "last_name"
     t.string "previous_name"
--- a/spec/system/referrals/user_adds_contact_details_email_spec.rb
+++ b/spec/system/referrals/user_adds_contact_details_email_spec.rb
@@ -1,0 +1,101 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+RSpec.feature "Contact details" do
+  scenario "User submits contact details for the referred person" do
+    given_the_service_is_open
+    and_the_employer_form_feature_is_active
+    and_i_have_an_existing_referral
+    when_i_visit_the_referral_summary
+    and_i_click_on_contact_details
+    then_i_see_the_personal_email_address_page
+
+    when_i_select_yes_without_email
+    and_i_press_continue
+    then_i_see_a_missing_email_error
+
+    when_i_select_yes_with_an_invalid_email
+    and_i_press_continue
+    then_i_see_an_invalid_email_error
+
+    when_i_select_yes_with_a_valid_email
+    and_i_press_continue
+    then_i_get_redirected_to_the_referral_summary
+
+    and_i_click_on_contact_details
+    when_i_select_no
+    and_i_press_continue
+    then_i_get_redirected_to_the_referral_summary
+  end
+
+  private
+
+  def given_the_service_is_open
+    FeatureFlags::FeatureFlag.activate(:service_open)
+  end
+
+  def and_the_employer_form_feature_is_active
+    FeatureFlags::FeatureFlag.activate(:employer_form)
+  end
+
+  def and_i_have_an_existing_referral
+    @referral = Referral.create!
+  end
+
+  def when_i_visit_the_referral_summary
+    visit edit_referral_path(@referral)
+  end
+
+  def and_i_click_on_contact_details
+    click_link "Contact details"
+  end
+
+  def then_i_see_the_personal_email_address_page
+    expect(page).to have_current_path(
+      "/referrals/#{@referral.id}/contact-details/email"
+    )
+    expect(page).to have_title(
+      "Do you know the personal email address of the person you are referring?"
+    )
+    expect(page).to have_content(
+      "Do you know the personal email address of the person you are referring?"
+    )
+  end
+
+  def when_i_select_yes_without_email
+    choose "Yes", visible: false
+  end
+
+  def and_i_press_continue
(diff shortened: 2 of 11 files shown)


spec/system/referrals/user_adds_contact_details_email_spec.rb
```diff
@@ -0,0 +1,101 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+RSpec.feature "User adds the contact details email to a referral" do
+  scenario "Views referral summary sections" do
```
malcolmbaig, line 5:
```suggestion
RSpec.feature "Contact details" do
  scenario "User submits contact details for the referred person" do
```

- with an update to the filename to match the scenario
- and then this spec could be used for all further contact detail steps in the flow

Otherwise spec structure below looks 👌🏼

app/controllers/referrals_controller.rb
```diff
@@ -34,7 +34,7 @@ def delete
 
   def referral
     # TODO: This needs integration with current_user check
```
malcolmbaig, line 36: 👍🏼

app/controllers/referrals_controller.rb
```diff
@@ -34,7 +34,7 @@ def delete
...
-    @referral ||= Referral.find(params[:id])
+    @referral ||= Referral.find_or_create_by(id: params[:id])
```
malcolmbaig, line 37: Should we be creating referrals here? I think it may be safer to limit creation to just one place in the app. If we don't find the referral by ID, then we can redirect them to a sensible starting point (the root?). cc: @steventux

app/controllers/referrals/contact_details_controller.rb
```diff
@@ -0,0 +1,33 @@
+module Referrals
+  class ContactDetailsController < ReferralsController
```
malcolmbaig, line 2: Perhaps this should be:

```suggestion
  class ContactDetails::EmailController < ReferralsController
```

with the actions named `edit`/`update`?

malcolmbaig (changes requested):
Looking good, a few comments left inline

app/controllers/referrals_controller.rb
vassyz: This was the concern I raised in the standup last week, where the referral should get created. The code above is from @steventux 's PR.

steventux: This `find_or_create_by` is here for convenience at the moment while we build out these forms. I wouldn't expect this to remain as is for several reasons.
I had imagined we would create a referral off the back of the eligibility screener as the two records may be associated.
Also a Referral should be associated with a User account.

app/controllers/referrals/contact_details_controller.rb
steventux: I was wondering about this approach on the personal details forms. Are we better off with multiple slim controllers and fewer actions? I'm happy either way but will make similar changes [here](https://github.com/DFE-Digital/refer-serious-misconduct/pull/94) for consistency.

app/controllers/referrals_controller.rb
malcolmbaig: 👍🏼 Makes sense. I'll add a task to the User accounts work to go through this and make sure referral querying is scoped to current user.

tvararu: :+1: thanks @malcolmbaig

app/controllers/referrals/contact_details_controller.rb
tvararu: I think multiple slim controllers with idiomatic `edit/update` actions, plus each one with its own `FormModel` makes sense :+1:

steventux (approved):
A question about the boolean db field (more out of ignorance). Otherwise all looks good to me.

app/forms/referrals/contact_details_email_form.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    attr_accessor :referral, :email_address
+    attr_reader :email_known
+
+    EMAIL_REGEX = ValidForNotifyValidator::EMAIL_REGEX
+
+    validates :referral, presence: true
+    validates :email_known, inclusion: { in: [true, false] }
+    validates :email_address, presence: true, if: -> { email_known }
+    validates_format_of :email_address,
```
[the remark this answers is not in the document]
vassyz: Thanks! That's a lot better.

db/schema.rb
```diff
@@ -22,6 +22,8 @@
     t.string "teaching_in_england"
     t.string "unsupervised_teaching"
     t.string "is_teacher"
+    t.boolean "email_known"
```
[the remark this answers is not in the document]
steventux: Makes sense, I think persisting is right

malcolmbaig (approved):

PR #105, opened 2022-11-01
Add a performance dashboard

felixclack:
We want to provide a snapshot of how the service is being used.

The design and metrics for this page have been lifted from Find.

I assume we will iterate on this and display more appropriate metrics as
we have need.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

Screenshot 2022-11-01 at 9 02 52 am

### DIFF
--- a/app/models/performance_stats.rb
+++ b/app/models/performance_stats.rb
@@ -1,0 +1,184 @@
+# frozen_string_literal: true
+class PerformanceStats
+  include ActionView::Helpers::NumberHelper
+
+  LAUNCH_DATE = Date.new(2022, 10, 1).beginning_of_day.freeze
+
+  def daily_percentiles
+    @daily_percentiles ||=
+      last_n_days.map do |day|
+        percentiles = percentiles_by_day[day] || [0, 0, 0]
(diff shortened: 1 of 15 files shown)


app/models/performance_stats.rb
```diff
@@ -0,0 +1,184 @@
+# frozen_string_literal: true
+class PerformanceStats
+  include ActionView::Helpers::NumberHelper
+
+  LAUNCH_DATE = Date.new(2022, 10, 1).beginning_of_day.freeze
```
tvararu, line 5:
```suggestion
  # TODO: Update when we go live
  LAUNCH_DATE = Date.new(2022, 10, 1).beginning_of_day.freeze
```

tvararu (approved):
Minor: Since this PR contains a lot of code copied from another project, it would have been easier to review if we had separate commits for the imported bits so they can just get a gloss-over.

I mostly looked through app/models/performance_stats.rb and the new filters in app/models/eligibility_check.rb and didn't spot anything out of the ordinary. LGTM

PR #118, opened 2022-11-03
Your name page

felixclack:
The first piece of information we collect on the referrer is their name.

An assumption here is that simply the presence of text is enough to
proceed with the referral.

In the designs, the referrer details screen has a way to indicate
whether the flow has been completed or not. This is left to a future PR
to avoid this one getting overly complex.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

Screenshot 2022-11-03 at 10 30 00 am

Screenshot 2022-11-03 at 10 01 43 am


steventux (approved):
One question about the scoping of `Referral.find`, otherwise all looks good.

PR #120, opened 2022-11-03
Fix incorrect fields in the referrals table

vassyz:
### Context
A broken migration created some unnecessary fields in the database. 

The migration will change the data type of the `email_known` field from `string` to `boolean`  and remove two others fields (`string` and `boolean`). One factory was also annotated on migration.

### Guidance to review
I had to format the migration in a certain way to avoid the rubocop lint errors.

### Checklist

- [ ] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/db/migrate/20221026143109_add_teacher_contact_details_email_to_referrals.rb
+++ b/db/migrate/20221026143109_add_teacher_contact_details_email_to_referrals.rb
@@ -1,8 +1,8 @@
 class AddTeacherContactDetailsEmailToReferrals < ActiveRecord::Migration[7.0]
   def change
     change_table :referrals, bulk: true do |t|
-      t.string :email_known, :boolean
-      t.string :email_address, :string, limit: 256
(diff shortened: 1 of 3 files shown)


spec/factories/eligibility_checks.rb
```diff
@@ -1,3 +1,17 @@
+# == Schema Information
```
steventux, line 1: Is this auto-generated when running migrations? Seems unrelated but harmless.

steventux (approved):
:rocket:

spec/factories/eligibility_checks.rb
steventux: Maybe it's better to commit it here then and we won't see it in subsequent diffs.

vassyz: Rebased form main and it's gone, already committed by somebody else 😄

PR #123, opened 2022-11-04
Collect the referrer's telephone number

felixclack:
We want a way of being able to contact the referrer about their
referral.

The design history page for this question is
https://teacher-misconduct.herokuapp.com/report/your-details/telephone.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/referrals/referrer_phone_controller.rb
+++ b/app/controllers/referrals/referrer_phone_controller.rb
@@ -1,0 +1,30 @@
+module Referrals
+  class ReferrerPhoneController < ApplicationController
+    def edit
+      @referrer_phone_form = ReferrerPhoneForm.new(referral: current_referral)
+    end
+
+    def update
+      @referrer_phone_form =
+        ReferrerPhoneForm.new(
+          referrer_phone_form_params.merge(referral: current_referral)
+        )
+      if @referrer_phone_form.save
+        redirect_to referral_referrer_details_path(current_referral)
+      else
+        render :edit
+      end
+    end
+
+    private
+
+    def current_referral
+      @current_referral ||= Referral.find(params[:referral_id])
+    end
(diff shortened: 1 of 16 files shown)


app/controllers/referrals/referrer_phone_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+        render :edit
+      end
+    end
+
+    private
+
+    def current_referral
+      @current_referral ||= Referral.find(params[:referral_id])
+    end
+    helper_method :current_referral
```
malcolmbaig, line 24: 👌🏼 

Worth adding this to ApplicationController to encourage usage elsewhere?

app/models/referrer.rb
```diff
@@ -4,6 +4,7 @@
 #
```
malcolmbaig, line 4: General comment: something to think about shortly, when I start hooking up the User accounts to the employer form — does it make sense to replace this model with the User one?

spec/system/referrals/about_you_spec.rb
```diff
@@ -106,4 +127,8 @@ def when_i_click_save_and_continue
   def when_i_enter_my_name
     fill_in "What is your name?", with: "Test Name"
   end
+
+  def when_i_enter_my_telephone_number
+    fill_in "What is your telephone number?", with: "01234567890"
+  end
 end
```
malcolmbaig, line 134: Suggestions, for consistency with the styleguide:

- Replace describe/it with feature/scenario
- A scenario description of "user provides their details"
- Renaming the file to match the scenario description

malcolmbaig (approved):
LG, some suggestions left inline 👍🏼

PR #134, opened 2022-11-08
Add the ability to specify if the form is complete

felixclack:
When filling in the about you section of the referral, we offer the
ability for someone to indicate whether they have completed a section or
not.

Assumptions:

* no validation required on the complete flag because all fields have a
  presence validation on them anyway
* at some point the presence of the complete flag will be used to
  validate submitting the referral

### Checklist

- [ ] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

Screenshot 2022-11-08 at 12 01 42 pm

### DIFF
--- a/db/migrate/20221107130734_add_completed_at_to_referrer.rb
+++ b/db/migrate/20221107130734_add_completed_at_to_referrer.rb
@@ -1,0 +1,5 @@
+class AddCompletedAtToReferrer < ActiveRecord::Migration[7.0]
+  def change
+    add_column :referrers, :completed_at, :timestamp
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 20_221_107_114_023) do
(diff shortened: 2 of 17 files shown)


db/schema.rb
```diff
@@ -66,6 +66,7 @@
     t.datetime "updated_at", null: false
     t.string "phone"
     t.string "job_title"
+    t.datetime "completed_at", precision: nil
```
steventux, line 69: I think I prefer the presence of a timestamp over a boolean, which I've used for the personal details section. I'll amend for consistency.

db/migrate/20221107130734_add_completed_at_to_referrer.rb
```diff
@@ -0,0 +1,5 @@
+class AddCompletedAtToReferrer < ActiveRecord::Migration[7.0]
+  def change
+    add_column :referrers, :completed_at, :timestamp
```
steventux, line 3: The referral summary page can display possible 3 states for a section.

- Not started yet
- Incomplete
- Completed

https://github.com/DFE-Digital/refer-serious-misconduct/blob/main/app/forms/referral_form.rb#L100-L105

I think presence of a timestamp limits us to 2 states.

steventux (commented):
A question about section status which might influence the use of a timestamp, happy to discuss.

PR #142, opened 2022-11-09
Associate users and referrals

malcolmbaig:
### Context

Users can now sign in, but the relationship between them and their referrals needs defining. Among other things, this allows us to ensure referral database queries are correctly scope to the user that owns them.

Broadly, this PR contains changes which:

- establish the model relationship
- scope referral querying in the controllers
- implement the interstitial page just prior to viewing the referral summary (the continue button here is the thing which creates the actual referral record). There's more work to be done to connect the eligibility screener flow to this page.

See commits for a more detailed summary.

### Guidance to review

Best way to test manually is to navigate to /referrals/new

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/controllers/referrals_controller.rb
+++ b/app/controllers/referrals_controller.rb
@@ -1,5 +1,6 @@
 class ReferralsController < Referrals::BaseController
   before_action :check_employer_form_feature_flag_enabled
+  before_action :redirect_to_referral_if_exists, only: %i[new create]
 
   def new
   end
@@ -39,6 +40,12 @@ class ReferralsController < Referrals::BaseController
   end
 
   private
+
+  def redirect_to_referral_if_exists
+    latest_referral = current_user.latest_referral
+
+    redirect_to referral_path(latest_referral) if latest_referral
+  end
 
   def referral
     @referral ||= current_user.referrals.find(params[:id])
--- a/app/controllers/referrals_controller.rb
+++ b/app/controllers/referrals_controller.rb
@@ -2,6 +2,14 @@ class ReferralsController < Referrals::BaseController
   before_action :check_employer_form_feature_flag_enabled
 
   def new
+  end
+
+  def create
+    referral = current_user.referrals.create
+    redirect_to referral_path(referral)
+  end
+
+  def show
     @referral_form = ReferralForm.new(referral:)
   end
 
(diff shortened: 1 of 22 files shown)


app/controllers/referrals_controller.rb
```diff
@@ -2,6 +2,14 @@ class ReferralsController < Referrals::BaseController
   before_action :check_employer_form_feature_flag_enabled
 
   def new
+  end
+
+  def create
+    referral = current_user.referrals.create
+    redirect_to referral_path(referral)
+  end
```
malcolmbaig, line 10: A consequence of this is that it's trivial to hit the back button/continue repeatedly to create a bunch of new referral records. Not a major issue, IMHO, because that would be the case in any system that allows the user to create new records via the UI. 

However, the MVP design is quite focused around the idea of creating a single referral, and doesn't really outline how we'd handle subsequent ones. Specifically, there's no concept of an index page where you'd see and manage these.

Should we temporarily hack this by redirecting if a referral already exists?

tvararu: Yes

app/models/referral.rb
```diff
@@ -26,9 +26,19 @@
...
+#
+#  index_referrals_on_user_id  (user_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (user_id => users.id)
 #
 class Referral < ApplicationRecord
   has_one :referrer, dependent: :destroy
+  belongs_to :user
```
steventux, line 41: Have we consciously separated the `User` from the `Referrer`? 
If I'm the user logging in to make a referral, aren't I the referrer? 
Asking because getting total referring user information requires joining `Referrer`, `Referral` and `User`. This might not be particularly significant.

steventux: I can definitely see the sense in keeping the model concerned with session activity distinct from the model concerned with details, I think leaving the data model as-is is more flexible.

steventux (approved):

app/models/referral.rb
vassyz: I think it's safer to keep them separated. It makes more sense to move the `Referrer` details into `Referral`.

PR #143, opened 2022-11-09
Move data validations to module

vassyz:
### Context

We need a was of re-using the date validation for day/month/year date inputs. I moved the exisitn code from the `AgeForm` to a concern.

### Checklist

- [ ] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/forms/concerns/validated_date.rb
+++ b/app/forms/concerns/validated_date.rb
@@ -1,0 +1,103 @@
+# frozen_string_literal: true
+module ValidatedDate
+  extend ActiveSupport::Concern
+
+  included do
+    def validated_date(date_params:, attribute:, date_of_birth: false)
+      date_fields = [
+        date_params["#{attribute}(1i)"],
+        date_params["#{attribute}(2i)"],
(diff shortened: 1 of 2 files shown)


steventux:
I also had a look into how easy this would be to refactor and the knottiness seems to be around multiple things happening in one place:

1. Conversion - we don't use Rails' in-built date fields to Date conversion, but instead attempt to allow months as words and numbers as words etc.
2. Validation - This should be simple enough to extract into its own validator but should only validate.
3. Assignment - the same piece of code assigns back onto the model attribute somewhere in the validation logic.

I think picking these 3 things apart would help with what the code is trying to do here. 
`AgeForm` was copied over from find-a-lost-trn and adapted slightly but I think it might be worth a proper refactor if the goal is to make a shared value converter & validator.

steventux (approved):
As mentioned I think this could ultimately do with splitting up into conversion and validation but as we need to re-use for other date fields this is a good start. :+1:

PR #149, opened 2022-11-10
Put user authentication behind a feature flag

malcolmbaig:
### Context

We need to be able to switch user auth on/off so that a simpler,  screener-only version of the service can be shipped as an MVP.

Add a `user_accounts` feature flag, and a set of conditional checks around the codebase.

When inactive, the service simply consists of the eligibility screener and a final page pointing towards the appropriate offline forms.

When active, the user auth flow kicks in immediately after the start page and connects the eligibility screener to the online referral form.

### Guidance to review

- Test from root path with feature on and off

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -1,12 +1,26 @@
 class PagesController < ApplicationController
+  def start
+    @start_now_path =
+      if FeatureFlags::FeatureFlag.active?(:user_accounts)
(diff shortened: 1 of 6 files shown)


steventux (approved):
:rocket:

tvararu (approved):
:raised_hands:

PR #151, opened 2022-11-10
Set up SSH on azure

steventux:
### Context

Azure has specific requirements for SSH access.
See https://learn.microsoft.com/en-us/azure/app-service/configure-custom-container?pivots=container-linux#enable-ssh

Adds necessary SSH dependencies and config for Azure container.

### Guidance to review

This PR is based on recent work [setting up SSH for Azure on Access Your Teaching Profile](https://github.com/DFE-Digital/access-your-teaching-profile/pull/44) 

### Checklist

- [ ] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -80,5 +80,14 @@ RUN apk add --no-cache libpq
 COPY --from=builder /app /app
 COPY --from=builder /usr/local/bundle/ /usr/local/bundle/
 
-CMD bundle exec rails db:migrate && \
+# SSH access specific to Azure
+# Install OpenSSH and set the password for root to "Docker!".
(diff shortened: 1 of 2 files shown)


app/controllers/referrals/allegation_details/details_controller.rb
```diff
@@ -0,0 +1,25 @@
+module Referrals
```
vassyz, line 1: Should these files be here?

steventux: No, removed

vassyz (approved):

PR #153, opened 2022-11-10
Open runtime container port 2222 for Azure SSH access

steventux:
### Context

Missed in https://github.com/DFE-Digital/refer-serious-misconduct/pull/151 ~we need to~ it's document we should open port 2222 for Azure SSH access.

Opens port 2222 on runtime container.

### Guidance to review

See https://github.com/DFE-Digital/access-your-teaching-profile/blob/main/Dockerfile#L85 for working example

### Checklist

- [ ] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


steventux:
Without merging this we currently have SSH access, unsure whether this is because it has been _hand configured_ or the docs are out of date.

tvararu (approved):

PR #154, opened 2022-11-10
Collect the organisation address

felixclack:
As part of collecting the referrer's organisation details, we want to
get the address.

This introduces a simple method of persisting the address. I opted to
keep it as fields of the organisation rather than creating a separate
object.

Assumptions:

* an address doesn't need to be verified in any way
* storing duplicate addresses across the service is fine

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

Screenshot 2022-11-11 at 12 30 54 pm
Screenshot 2022-11-11 at 12 30 48 pm

### DIFF
--- a/app/controllers/referrals/organisation_name_controller.rb
+++ b/app/controllers/referrals/organisation_name_controller.rb
@@ -11,13 +11,23 @@ module Referrals
           organisation_name_form_params.merge(referral: current_referral)
         )
       if @organisation_name_form.save
-        redirect_to referral_organisation_path(current_referral)
+        redirect_to(
+          if go_to_check_answers?
+            referral_organisation_path(current_referral)
+          else
+            edit_referral_organisation_address_path(current_referral)
+          end
+        )
       else
         render :edit
       end
     end
 
     private
+
+    def go_to_check_answers?
+      params["check_answers"] == "true"
+    end
 
     def organisation_name_form_params
       params.require(:organisation_name_form).permit(:name)
--- a/app/forms/organisation_address_form.rb
+++ b/app/forms/organisation_address_form.rb
@@ -1,0 +1,35 @@
+class OrganisationAddressForm
+  include ActiveModel::Model
+
+  attr_accessor :referral
+  attr_writer :street_1, :street_2, :city, :postcode
+
+  validates :city, presence: true
+  validates :postcode, presence: true
+  validates :referral, presence: true
+  validates :street_1, presence: true
+
+  def city
+    @city ||= organisation&.city
+  end
+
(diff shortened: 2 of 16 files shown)


app/models/organisation.rb
```diff
@@ -20,6 +24,19 @@
 class Organisation < ApplicationRecord
   belongs_to :referral
 
+  def address
```
vassyz, line 27: I put some similar code in a helper, as I don't think this should belong to a model: https://github.com/DFE-Digital/refer-serious-misconduct/blob/main/app/helpers/referrer_helper.rb

app/controllers/referrals/organisation_name_controller.rb
```diff
@@ -11,14 +11,24 @@ def update
           organisation_name_form_params.merge(referral: current_referral)
         )
       if @organisation_name_form.save
-        redirect_to referral_organisation_path(current_referral)
+        redirect_to(
+          if go_to_check_answers?
```
vassyz, line 15: Why is this needed?

app/forms/organisation_address_form.rb
```diff
@@ -0,0 +1,35 @@
+class OrganisationAddressForm
+  include ActiveModel::Model
+
+  attr_accessor :referral
+  attr_writer :street_1, :street_2, :city, :postcode
+
+  validates :city, presence: true
+  validates :postcode, presence: true
```
vassyz, line 8: I added a gem for validationg UK postcodes. maybe you want to use it here as well: https://github.com/DFE-Digital/refer-serious-misconduct/blob/main/app/forms/referrals/contact_details/address_form.rb#L21

spec/forms/organisation_address_form_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    context "when one of the validations is failing" do
+      let(:street_1) { "" }
+
+      it { is_expected.to be_falsy }
+    end
+  end
+
+  describe "#save" do
+    subject(:save) { form.save }
```
vassyz, line 38: I replaced this and those following `save` calls with a `before { form.save }`.

app/models/organisation.rb
vassyz: That method needs updating as you don't have all the fields. Also the `referral` argument name should be more general.

app/controllers/referrals/organisation_name_controller.rb
vassyz: Where is this requirement from? The pages I've worked on don't have this behaviour.

vassyz (approved):

PR #164, opened 2022-11-14
Add a check answers page for previous allegation

felixclack:
When a user adds details about previous allegations, we display a page
that allows them to review all the details they provided and confirm if
they have finished the section.

This introduces that page and allows us to iterate on the previous
allegation flow.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

Screenshot 2022-11-14 at 10 59 03 am

### DIFF
--- a/app/controllers/referrals/previous_misconduct_controller.rb
+++ b/app/controllers/referrals/previous_misconduct_controller.rb
@@ -1,0 +1,26 @@
+class Referrals::PreviousMisconductController < Referrals::BaseController
+  def show
+    @previous_misconduct_form =
+      PreviousMisconductForm.new(referral: current_referral)
+  end
+
+  def update
+    @previous_misconduct_form =
+      PreviousMisconductForm.new(
+        complete: previous_misconduct_params[:complete],
+        referral: current_referral
+      )
+
+    if @previous_misconduct_form.save
+      redirect_to edit_referral_path(current_referral)
+    else
+      render :show
+    end
+  end
+
+  private
+
+  def previous_misconduct_params
+    params.require(:previous_misconduct_form).permit(:complete)
+  end
+end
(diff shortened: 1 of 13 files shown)


app/models/previous_allegation.rb
```diff
@@ -0,0 +1,31 @@
...
+#
+# Indexes
+#
+#  index_previous_allegations_on_referral_id  (referral_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (referral_id => referrals.id)
+#
+class PreviousAllegation < ApplicationRecord
```
steventux, line 19: If we are modelling previous allegations as an association on a referral would it be simpler to have an `Allegation` model and a state/scope for previous/current? 
Asking because if we do go down the associated model route we'll have two near identical db tables / models for both previous and current allegations.
I'm also curious as to where we draw the line with the referrals record, we did have a brief discussion when kicking off about minimizing associations, personally I think associating allegations works well but worth getting feedback from other devs @tvararu @malcolmbaig @vassyz WDYT?
(The PR looks good to me other than this Q)

app/forms/previous_allegation_form.rb
```diff
@@ -0,0 +1,28 @@
...
+  include ActiveModel::Model
+
+  attr_accessor :referral
+  attr_writer :complete
+
+  validates :complete, inclusion: { in: %w[true false] }
+  validates :referral, presence: true
+
+  def complete
+    @complete || previous_allegation&.completed_at? || nil
```
vassyz, line 11: What is the purpose of ` || nil`?

vassyz: Can't you remove `let(:previous_allegation) { nil }` in your test? I don't think it makes a difference.

PR #170, opened 2022-11-15
Change sign in flash banner to :success

malcolmbaig:
Change colour from blue to green, more in line with GDS design system.

https://trello.com/c/qDvX8DHU/939-rsm-user-accounts-integrate-with-eligibility-checker-and-employer-form#comment-637214bf1f5a420092bd7d1e

Screenshot_20221115_110440


steventux (approved):

tvararu (approved):
I think even more in line would be to have it be two-thirds width; can be done as a follow-up.

PR #173, opened 2022-11-15
Add review app

NickGraham101:
### Context

To test changes, we require the option to deploy temporary review apps.

- Changes to terraform and addition of workflows to deploy review apps
- Addition of workflow to delete review apps
- Remove `prevent_destroy` lifecycle metadata that protects Azure Storage from accidental deletion by terraform.  This provides limited value (any service or user other than terraform can still delete the storage account) and prevents the creation of a temporary storage account for the review app

### Guidance to review

- To test the delete process run `gh workflow run "Delete Review App" --ref=add-review-app -f pr_number=173` locally
- To test the review app deployment process rerun the last Build and Action triggered by this branch
- Tested changes against dev environment [here](https://github.com/DFE-Digital/refer-serious-misconduct/actions/runs/3488756326)

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -31,6 +31,36 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           registry: ghcr.io
 
+  deploy_review_app:
+    name: Deploy to review environment
+    runs-on: ubuntu-latest
+    if: contains(github.event.pull_request.labels.*.name, 'deploy')
+    concurrency: deploy_review_${{ github.event.pull_request.number }}
+    needs: [build_image]
+    environment:
+      name: review
+      url: ${{ steps.deploy.outputs.environment_url }}
+    outputs:
+      environment_url: ${{ steps.deploy.outputs.environment_url }}
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: ./.github/actions/deploy-environment
+        id: deploy
+        with:
+          environment_name: review
+          image_name_tag: ${{ needs.build_image.outputs.image_name_tag }}
+          image_tag: ${{ github.sha }}
+          azure_credentials: ${{ secrets.AZURE_CREDENTIALS }}
+          site_up_retries: 150
+
+      - uses: ./.github/actions/smoke-test
+        id: smoke-test
+        with:
+          environment: review
+          azure_credentials: ${{ secrets.AZURE_CREDENTIALS }}
+
   deploy:
     name: Deploy to ${{ matrix.environment }} environment
     runs-on: ubuntu-latest
--- a/.github/workflows/delete-review-app.yml
+++ b/.github/workflows/delete-review-app.yml
@@ -1,0 +1,130 @@
+name: Delete Review App
+
+on:
+  pull_request:
+    types: [closed]
+    branches: [main]
+  workflow_dispatch:
+    inputs:
+      pr_number:
+        description: PR number of review app to delete
+        required: true
+        type: string
+jobs:
+  delete-review-app:
+    name: Delete Review App ${{ github.event.pull_request.number }}
+    concurrency: deploy_review_${{ github.event.pull_request.number }}
+    runs-on: ubuntu-latest
+    environment: review
+    steps:
+      - uses: actions/checkout@v3
+      - name: Extract configuration from tfvars
+        id: config
+        run: |
+          if [ ${{ github.event_name }} == 'workflow_dispatch' ]; then
+            PR_NUMBER=${{ github.event.inputs.pr_number }}
+          else
+            PR_NUMBER=${{ github.event.pull_request.number }}
+          fi
+
+          if [ -z "$PR_NUMBER" ]; then
+            echo "::error ::Failed to extract PR_NUMBER"
+            exit 1
+          fi
+
+          RESOURCE_PREFIX=$(jq -r '.resource_prefix' $TFVARS)
+          APP_RESOURCE_GROUP_NAME=$RESOURCE_PREFIX-review-pr-$PR_NUMBER-rg
+          STORAGE_ACCOUNT_NAME=$(jq -r '.storage_account_name' $TFVARS)
+          TERRAFORM_VERSION=$(awk '/{/{f=/^terraform/;next}f' terraform.tf | grep -o [0-9\.]*)
+          DEV_TFVARS=workspace_variables/dev.tfvars.json
+          TF_RESOURCE_GROUP_NAME=$(jq -r '.resource_group_name' $DEV_TFVARS)
+          TF_STATE_CONTAINER=$(awk '/{/{f=/backend/;next}f' terraform.tf | grep -o "\"[a-z\-]*\"" | tr -d \")
+
+          if [ -z "$APP_RESOURCE_GROUP_NAME" ]; then
+            echo "::error ::Failed to extract app_resource_group_name from $TFVARS"
+            exit 1
+          fi
+
+          if [ -z "$STORAGE_ACCOUNT_NAME" ]; then
+            echo "::error ::Failed to extract storage_account_name from $TFVARS"
+            exit 1
+          fi
+
+          if [ -z "$TERRAFORM_VERSION" ]; then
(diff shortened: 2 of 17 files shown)


NickGraham101:
Testing delete-review-app workflow

NickGraham101:
[Implement a process for deploying review apps in Azure](https://trello.com/c/VQyOErf1/727-implement-a-process-for-deploying-review-apps-in-azure)

.github/workflows/delete-review-app.yml
```diff
@@ -0,0 +1,106 @@
...
+    inputs:
+      pr_number:
+        description: PR number of review app to delete
+        required: true
+        type: string
+jobs:
+  delete-review-app:
+    name: Delete Review App ${{ github.event.pull_request.number }}
+    concurrency: deploy_review_${{ github.event.pull_request.number }}
+    #if: ${{ contains(github.event.pull_request.labels.*.name, 'deploy') }} ##TO DO: is this the best way, alts either labels or run every close and test for review app
```
saliceti, line 17: We have this condition in the other services. It cleanly skips the job when there is no review app, otherwise it would error which is not so nice

terraform/app.tf
```diff
@@ -5,7 +5,7 @@ locals {
       ApplicationInsights__ConnectionString = azurerm_application_insights.insights.connection_string
       DATABASE_URL                          = "postgres://postgres@${local.postgres_server_name}.postgres.database.azure.com:5432"
       DATABASE_PASSWORD                     = local.infrastructure_secrets.POSTGRES_ADMIN_PASSWORD
-      HOSTING_DOMAIN                        = "https://${var.domain}"
+      HOSTING_DOMAIN                        = var.domain != null ? "https://${var.domain}" : ""
```
saliceti, line 8: Don't we always need it? it could be set to the app service domain by default

terraform/monitor-diagnostics.tf
```diff
@@ -1,4 +1,5 @@
-resource "azurerm_monitor_diagnostic_setting" "getanid-keyvault-diagnostics" {
+resource "azurerm_monitor_diagnostic_setting" "rsm-keyvault-diagnostics" {
+  count    = var.environment_name == "review" ? 0 : 1
```
saliceti, line 2: Instead of hard coding the env name, could we create a feature flag variable such as `enable_keyvault_diagnostics`? It would default to true and we would set it false for review apps

saliceti: Or reuse variable `keyvault_logging_enabled`?

.github/workflows/delete-review-app.yml
NickGraham101: My intention was to ensure that review apps are deleted even if devs remove the deploy label before closing it.  That had been mentioned as a problem in standup recently.  I think I can do that in a way that doesn't result in the workflow run failing

.github/workflows/build-and-deploy.yml
```diff
@@ -31,6 +31,36 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           registry: ghcr.io
 
+  deploy_review_app:
+    name: Deploy to review environment
+    runs-on: ubuntu-latest
+    if: contains(github.event.pull_request.labels.*.name, 'deploy')
+    concurrency: deploy__review_${{ github.event.pull_request.number }}
```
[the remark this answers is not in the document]
NickGraham101: No, removed it

.github/workflows/delete-review-app.yml
```diff
@@ -0,0 +1,106 @@
+name: Delete Review App
+
+on:
+  pull_request:
+    types: [closed]
+    branches: [main]
+  workflow_dispatch:
```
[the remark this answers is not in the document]
NickGraham101: Makes it easier to delete any review apps that left behind imo but it isn't essential

.github/workflows/delete-review-app.yml
```diff
@@ -0,0 +1,106 @@
...
+        shell: bash
+        env:
+          TFVARS: workspace_variables/review.tfvars.json
+        working-directory: terraform
+
+      - uses: Azure/login@v1
+        with:
+          creds: ${{ secrets.azure_credentials }}
+
+      # get TFSTATE-CONTAINER-ACCESS-KEY
```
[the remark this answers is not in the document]
NickGraham101: Deleted along with similar comment in deploy action

.github/workflows/delete-review-app.yml
```diff
@@ -0,0 +1,106 @@
...
+          echo "::add-mask::$TFSTATE_CONTAINER_ACCESS_KEY"
+          echo "TFSTATE_CONTAINER_ACCESS_KEY=$TFSTATE_CONTAINER_ACCESS_KEY" >> $GITHUB_ENV
+        shell: bash
+
+      - uses: hashicorp/setup-terraform@v2
+        with:
+          terraform_version: ${{ ENV.terraform_version }}
+          terraform_wrapper: false
+
+      ##TO DO: parameterise container name
```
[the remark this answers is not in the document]
NickGraham101: That's what I meant 🙂 
Uses env var now

terraform/app.tf
NickGraham101: Doesn't look like it, the review app is passing the smoke tests without it.  Adding it in would mean hardcoding the value as there's a chicken and egg situation trying to pass in the app service's hostname to the app services app settings.  All the permanent environments have a domain, it's only the review environments that don't

terraform/review-app.tf
```diff
@@ -0,0 +1,6 @@
+resource "azurerm_resource_group" "review_app_rg" {
```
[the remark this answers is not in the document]
NickGraham101: We create the resource groups for other environments with ARM because terraform requires storage and key vault to already exist.  Review uses the dev storage and key vault

terraform/monitor-diagnostics.tf
NickGraham101: Good spot, have reused `keyvault_logging_enabled`

NickGraham101: Successfully tested this change and all others [here](https://github.com/DFE-Digital/refer-serious-misconduct/actions/runs/3495933319)

.github/workflows/delete-review-app.yml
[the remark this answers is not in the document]
saliceti: OK let's keep it, it doesn't add much complexity

terraform/app.tf
saliceti: Can't we predict the app service url from `local.rsm_app_name` ?

terraform/review-app.tf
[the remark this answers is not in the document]
saliceti: And we don't want to deploy review apps to the dev RG to avoid polluting it? Makes sense.
Can we use a feature flag variable here then? such as: `create_env_resource_group: true/false` instead of hard coding the env name

.github/workflows/delete-review-app.yml
```diff
@@ -72,7 +78,15 @@ jobs:
           terraform_version: ${{ ENV.terraform_version }}
           terraform_wrapper: false
 
+      - name: Check resource group exists
+        run: |
+          GROUP=$(az group show --name ${{ env.app_resource_group_name }} --query name)
```
saliceti, line 83: The above would be true if the command returns any non empty string
Can we use `az group exists --name ${{ env.app_resource_group_name }}` instead? This seems more reliable

.github/workflows/delete-review-app.yml
saliceti: OK - Discussed IRL

.github/workflows/delete-review-app.yml
NickGraham101: Good shout, done

terraform/app.tf
NickGraham101: We can, that's done

terraform/review-app.tf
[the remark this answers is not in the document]
NickGraham101: Sounds good, that's done

saliceti (approved):

PR #176, opened 2022-11-15
Employer form - Allegation section

steventux:
### Context

Referring misconduct requires the user to give details of the allegation

https://trello.com/c/UtZ0yWmd/948-employer-form-allegation-how-to-tell-us-page

This PR encompasses the whole allegation section.

#### Details:

image

#### DBS:

image

#### Check answers:

image

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


app/forms/referrals/allegation/confirm_form.rb
```diff
@@ -0,0 +1,39 @@
...
+      end
+
+      def allegation_not_incomplete
+        if referral.allegation_format == "incomplete"
+          errors.add(:base, :allegation_incomplete)
+        end
+      end
+
+      def allegation_details
+        if referral.allegation_upload.attached?
```
vassyz, line 29: There is a small issue with this, but as the form format might change, it's probably not worth putting too much effort into it. If you upload a file, and then decide to add text details instead, the file won't be deleted. In the table you'll see the filename because `referral.allegation_upload.attached?` will return true. It's the same issue if you do it the other way around. When you add text details and then go back and upload a file instead, it doesn't delete the text.

vassyz (approved):

app/forms/referrals/allegation/confirm_form.rb
steventux: :+1: Ah good spot! I'll take a look and see if there's a workaround for this.

PR #177, opened 2022-11-15
Remove annotate gem

vassyz:
### Context

The annotations weren't used by the team and because of the merge conflicts/expanding models we decided to remove the gem.

Removes the gem, rake task and the annotations from models, factories and specs.

### Checklist

- [ ] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


db/schema.rb
```diff
@@ -49,28 +49,30 @@
...
-    t.boolean "personal_details_complete"
     t.boolean "phone_known"
     t.string "phone_number"
-    t.boolean "age_known"
+    t.boolean "personal_details_complete"
     t.boolean "contact_details_complete"
     t.bigint "user_id", null: false
+    t.boolean "age_known"
+    t.boolean "role_start_date_known"
+    t.date "role_start_date"
```
malcolmbaig, line 75: I think these schema diffs shouldn't form part of this PR, as they don't relate to the described change.

vassyz: It was a mistake, I'll remove it. It's a bit annoying that we all have a different DB structure locally.

malcolmbaig (approved):

PR #182, opened 2022-11-16
Employer form - Evidence Start / Check answers page

steventux:
### Context

Users can upload and categorise evidence when making a referral.
https://trello.com/c/ZAQub52c/978-rsm-employer-form-evidence-check-answers-page

Adds the start and end of the evidence section flow. If the user answers 'No' to uploading evidence they are taken directly to the confirmation page.

image

Evidence upload steps to follow in another PR as this will be fairly substantial.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


app/forms/referrals/evidence/confirm_form.rb
```diff
@@ -0,0 +1,24 @@
...
+
+      def evidence_details_complete=(value)
+        @evidence_details_complete = ActiveModel::Type::Boolean.new.cast(value)
+      end
+
+      def save
+        return false if invalid?
+
+        referral.update(evidence_details_complete:)
+        true
```
vassyz, line 20: I saw this `true` returned from the `save` method in multiple places, but I'm not sure it's needed.

steventux: I guess it's not, `update` returns the object being updated which isn't strictly speaking `true` but that's perhaps a reason to use `truthy` assertions in specs rather than do this.

app/views/referrals/evidence/start/edit.html.erb
```diff
@@ -0,0 +1,51 @@
...
+
+    <p>Do not redact any of the documents. This will make it difficult to consider the evidence. All
+    information given is treated in strict confidence.</p>
+
+    <%= form_with model: @evidence_start_form, url: referrals_update_evidence_start_path(current_referral), method: :put do |f| %>
+      <%= f.govuk_error_summary %>
+
+      <%= f.govuk_radio_buttons_fieldset(
+        :has_evidence,
+        legend: { text: "Do you have anything to upload?", size: "m" },
```
vassyz, line 43: I'm not sure if this will get updated in a future PR as I cant see it in the prototype, but based on the error message, should't it ask if you have evidence, or documents to upload? Anything sounds a bit unprofessioanl to me.

vassyz (approved):

app/views/referrals/evidence/start/edit.html.erb
steventux: I've just gone with the prototype design here but I guess we will review this in future design iterations.

app/forms/referrals/evidence/confirm_form.rb
steventux: I'll do a follow up PR to remove this and other occurances, I agree it would be better to remove the explicit return value and make this consistent with other form save methods.

steventux: https://github.com/DFE-Digital/refer-serious-misconduct/pull/183/commits/76a3a1f9f2dd8a5b7c9711aede9e25408153d5d2 in following PR

PR #190, opened 2022-11-17
Add OTP validations and guess limit

malcolmbaig:
### Context

For user sign in, currently we redirect the user back to the email page on an incorrect guess. We also don't limit the number of attempts.
https://trello.com/c/kV9YjeYJ

- Add validations around OTP length and correctness
- Limit the number of guesses to 5 before sending the user back to the email screen

### Guidance to review

Can be tested locally with the `user_accounts` feature flag switched on.

### Checklist

- [x] Attach to Trello card
- [ ] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/users/otp_controller.rb
+++ b/app/controllers/users/otp_controller.rb
@@ -6,23 +6,28 @@ class Users::OtpController < DeviseController
   end
 
   def new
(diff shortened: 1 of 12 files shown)


steventux (approved):
:rocket:

PR #198, opened 2022-11-21
Control access to the screener using a feature flag

felixclack:
We want to control access to the eligibility screener and redirect to a
more appropriate location in the case where we need to disable it.

This allows the GOVUK page to direct people to the screener and we can
control whether we present the new flow for people or simply redirect
them to the existing location on GOVUK.

The main benefit here is that we don't need to ask another team to
modify the redirect in the case of issues with the screener.

I considered using a `before_action` in a base controller to implement
the redirect, but then thought it might add a layer of indirection for
the maintainer of the code.

Using a constraint to restrict access to the screener routes in the
route configuration seemed the simplest and clearest for someone reading
the code.

Assumptions:
* the redirect is temporary, so I used a 307 status code
* the feature flag is temporary, so having it in a single file makes it
  easier to remove in the future
* the redirect location on GOVUK is static, so hardcoding it is fine vs
  an ENV

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -55,7 +55,10 @@ Rails.application.routes.draw do
   ) do
     get "/start",
         to:
-          redirect("https://www.gov.uk/report-teacher-misconduct", status: 307)
+          redirect(
+            "https://www.gov.uk/government/publications/teacher-misconduct-referral-form",
+            status: 307
+          )
   end
 
   root to: redirect("/start")
--- a/spec/system/eligibility_screener_feature_flag_spec.rb
+++ b/spec/system/eligibility_screener_feature_flag_spec.rb
@@ -20,7 +20,7 @@ RSpec.feature "Eligibility screener", type: :system do
 
   def then_i_see_the_govuk_page
     expect(page).to have_current_path(
-      "https://www.gov.uk/report-teacher-misconduct"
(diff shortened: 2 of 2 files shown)


tvararu (approved):

tvararu (changes requested):
Looks great, just hold off on merging for a second while I follow up with Daniel to see if the link I suggested is correct

config/routes.rb
```diff
@@ -25,25 +25,38 @@
...
+    get "/you-should-know", to: "pages#you_should_know"
+    get "/complete", to: "pages#complete"
+  end
+
+  constraints(
+    -> { !FeatureFlags::FeatureFlag.active?(:eligibility_screener) }
+  ) do
+    get "/start",
+        to:
+          redirect("https://www.gov.uk/report-teacher-misconduct", status: 307)
```
tvararu, line 58: I think this one is better:

```suggestion
          redirect("https://www.gov.uk/government/publications/teacher-misconduct-referral-form", status: 307)
```

tvararu (approved):
Ship it

config/routes.rb
tvararu: Yes, use this one @felixclack

PR #204, opened 2022-11-21
About their role forms - Add same organisation page

vassyz:
### Context

Employer can select if the person they are referring works in the same organisation.
An error message will be shown when submitting without selecting a value.

https://trello.com/c/c46KiqWB/987-about-their-role-forms-add-same-organisation-page
https://teacher-misconduct.herokuapp.com/report/teacher-role/school

Adds the following page:
Screenshot 2022-11-21 at 17 32 38

- Update the routes
- Generate a migration for adding the same organisation field to the referrals table.
- Add the form, including validation
- Link the form to the job title page

### Checklist

- [x] Attach to Trello card
- [] Rebased main
- [] Cleaned commit history

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -21,7 +21,8 @@ ActiveRecord::Schema[7.0].define(version: 2022_11_22_134453) do
     t.bigint "blob_id", null: false
     t.datetime "created_at", null: false
     t.index ["blob_id"], name: "index_active_storage_attachments_on_blob_id"
-    t.index ["record_type", "record_id", "name", "blob_id"], name: "index_active_storage_attachments_uniqueness", unique: true
+    t.index ["record_type", "record_id", "name", "blob_id"], name: "index_active_storage_attachments_uniqueness", 
+unique: true
   end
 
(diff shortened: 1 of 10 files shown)


vassyz:
[About their role forms: Add same organisation page](https://trello.com/c/c46KiqWB/987-about-their-role-forms-add-same-organisation-page)

db/schema.rb
```diff
@@ -75,45 +74,56 @@
     t.index ["referral_id"], name: "index_organisations_on_referral_id"
   end
 
+  create_table "referral_evidences", force: :cascade do |t|
```
steventux, line 77: This doesn't seem related to this PR, more likely an artifact of reviewing my PR?

steventux (commented):
Looks good but schema file appears to have more changes than introduced by the migration in this PR.

db/schema.rb
vassyz: Not sure how I missed that. I'll fix it.

steventux (approved):
:rocket:

PR #208, opened 2022-11-22
Expire User OTPs

malcolmbaig:
### Context

Generated OTPs now have a guess limit, but they don't currently expire. We can improve the security of this sign-in solution by giving each OTP a fixed period of validity.

- Add form and controller logic to check for OTP expiry, redirecting to the email screen if the code has expired
- Incidental refactoring of validations and nil handling (see commits for detail)

This code sets an expiry of 30 minutes. Is that too much / too little?

### Link to Trello card

https://trello.com/c/NhjzKEFY
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


app/controllers/users/otp_controller.rb
```diff
@@ -19,7 +19,13 @@ def new
   def create
     @otp_form = Users::OtpForm.new(user_params)
 
-    if @otp_form.valid?
+    if @otp_form.otp_expired?
```
steventux, line 22: This is a near identical set of steps to the `@otp_form.maximum_guesses?` conditional, the only difference is the text of the `flash[:warning]`, I wonder if it's worth either golfing them down into the else branch of `@otp_form.valid?` or wrapping the steps in a method.
Just a stylistic suggestion, not a blocker.

db/migrate/20221121142532_add_last_otp_created_at_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddLastOtpCreatedAtToUsers < ActiveRecord::Migration[7.0]
+  def change
+    add_column :users, :last_otp_created_at, :timestamp
```
steventux, line 3: Bit pedantic but do we need to prefix this field with `last_` ?

steventux (approved):
:+1: Looks good to me, two very opinionated suggestions. Not blockers.

app/controllers/users/otp_controller.rb
malcolmbaig: Thanks @steventux - there's some discussion around one of these error messages currently. Will pick this up in a follow-on PR once I know how we're proceeding.

steventux (approved):
:rocket:

PR #223, opened 2022-11-24
Redirect to start page from referrals flow

steventux:
### Context

If the employer form or user accounts features are not active we shouldn't allow users to access the referrals flow.
https://trello.com/c/OelDJYCH/993-redirect-away-from-referrals-flow-when-user-accounts-feature-flag-is-inactive

- Redirect to start path if either of the employer form or user accounts features are inactive.
- Refactor some common system spec steps into a support module.

### Guidance to review

The last commit isn't relevant to the redirecting, just DRYing up system spec steps

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/referrals/base_controller.rb
+++ b/app/controllers/referrals/base_controller.rb
@@ -4,6 +4,7 @@ module Referrals
     before_action :authenticate_user!,
                   if: -> { FeatureFlags::FeatureFlag.active?(:user_accounts) }
     before_action :redirect_if_feature_flag_disabled
+    before_action :redirect_referrals_requests_if_user_accounts_disabled
 
     private
 
@@ -29,6 +30,13 @@ module Referrals
 
     def go_to_check_answers?
       params["go_to_check_answers"] == "true"
+    end
+
+    def redirect_referrals_requests_if_user_accounts_disabled
+      return if request.path !~ %r{^/referral}
+      return if FeatureFlags::FeatureFlag.active?(:user_accounts)
+
+      redirect_to root_path
     end
   end
 end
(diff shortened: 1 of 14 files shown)


app/controllers/referrals/base_controller.rb
```diff
@@ -12,10 +17,15 @@ def current_referral
     end
     helper_method :current_referral
 
-    def redirect_if_feature_flag_disabled
-      return if FeatureFlags::FeatureFlag.active?(:employer_form)
+    def redirect_if_feature_flag_disabled(
+      feature_flag:,
+      message: nil,
```
vassyz, line 22: Why did you add the `message` and `path` arguments? They don't seem to be passed when the method gets called.

steventux: @vassyz I have remove this and reworked the PR.

app/controllers/referrals/base_controller.rb
```diff
@@ -30,5 +31,12 @@ def store_user_location!
     def go_to_check_answers?
       params["go_to_check_answers"] == "true"
     end
+
+    def redirect_referrals_requests_if_user_accounts_disabled
+      return if request.path !~ %r{^/referral}
```
malcolmbaig, line 36: Is there anything which inherits from this base controller, which doesn't match on the regex? Just thinking if inheriting from this controller is enough to warrant applying this conditional redirect.

spec/support/system/common_steps.rb
```diff
@@ -0,0 +1,51 @@
...
+  def when_i_click_save_and_continue
+    click_on "Save and continue"
+  end
+  alias_method :and_i_click_save_and_continue, :when_i_click_save_and_continue
+
+  def when_i_click_back
+    click_on "Back"
+  end
+  alias_method :and_i_click_back, :when_i_click_back
+end
```
malcolmbaig, line 51: 👌🏼

malcolmbaig (approved):
LG

PR #224, opened 2022-11-24
Amend referrals 'previous_misconduct_summary' column removal

malcolmbaig:
This is causing errors in the deployed environments because the column is already absent for some reason but the migration is being run. Try adding `if_exists` to get the migrations passing.


vassyz (approved):

malcolmbaig:
The error turned out to be from the review app work that's currently in progress. Closing PR.

malcolmbaig:
Reopening — the review app errors occurred because they were being run from scratch against a fresh db. The errors occurred because the original migration that added the column was deleted at some point. Going to merge this as it's an easy fix for what's currently a broken set of migrations.

PR #226, opened 2022-11-24
Add user sign in error interstitial

malcolmbaig:
### Context
Design review identified the need for some improvement in the user journey when the user fails to sign in and their OTP state is reset.

Add an error page that prompts the user to try signing in again when either of the following error states arise:
- Expired OTP
- Exhausted number of guesses

#### Expired

Screenshot_20221124_141832

#### Exhausted

Screenshot_20221124_141911

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/users/otp_controller.rb
+++ b/app/controllers/users/otp_controller.rb
@@ -20,9 +20,7 @@ class Users::OtpController < DeviseController
     @otp_form = Users::OtpForm.new(user_params)
 
     if @otp_form.otp_expired?
-      @otp_form.user.after_failed_otp_authentication
-      flash[:warning] = "Security code has expired. Try again."
-      redirect_to new_user_session_path
+      fail_and_retry(reason: :expired)
     elsif @otp_form.valid?
       self.resource = warden.authenticate!(auth_options)
       set_flash_message!(:success, :signed_in)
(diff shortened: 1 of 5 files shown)


vassyz:
I'm not sure if this is the best way of displaying errors. You are setting an URL param to decide what message to display in the view, but in the OTP controller you are already using flash messages for a valid request. Why not use flash messages for this as well?

malcolmbaig:
> I'm not sure if this is the best way of displaying errors. It sets an URL param to decide what message to display in the view, but the OTP controller is already using flash messages for a valid request. Why not use flash messages for this as well?

I'm sure the designers can speak on this in more detail, but generally we don't use flashes for error messaging. Most (all?) errors in the service currently are [inline validation errors](https://design-system.service.gov.uk/components/error-message/) or [error summaries](https://design-system.service.gov.uk/components/error-summary/) . The error states described in this PR aren't a good fit for either of those, as we redirect them away from the OTP form. 

This approach seems like a reasonable way to convey details of the error while conforming to some of the guidance in the Design System about when (and when not) to use [notification banners](https://design-system.service.gov.uk/components/notification-banner/).

PR #230, opened 2022-11-25
Display the confirmation page

felixclack:
After submitting the referral, we want to display a confirmation page to
the referrer with instructions about what to expect to happen next.

In order to be able to submit the referral, all sections need to be
complete.

There is some difference in the values of the status fields for each
section. The validation on the form is expecting it to be `complete` but
the sections themselves use either `complete` or `completed`.

The correct styling of the label is triggered by the value `completed`,
which is also the word used in the designs.

This change ensures we use a consistent status value for all sections.

I made a decision to use the `edit` action to represent the summary page
in edit mode and the `show` action will be the review page. Currently,
they both render the same view.

### Link to Trello card

https://trello.com/c/DnEdcWuV/997-confirmation-page-https-teacher-misconductherokuappcom-report-submit-confirmation-already-implemented-needs-wiring-up

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history
Screenshot 2022-11-25 at 1 22 52 pm


spec/system/user_submits_referral_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    @user = create(:user)
+    sign_in(@user)
+  end
+
+  def and_i_click_review_and_send
+    click_on "Review and send"
+  end
+
+  def and_i_have_an_incomplete_referral
+    @referral = create(:referral, :complete, user: @user)
```
malcolmbaig, line 30: This data setup is a little confusing - there's a bit of mismatch between the method name and what it looks like the factory call is doing. I'd expect a `:complete` trait to give me back a complete referral — including the associations we create below in the `when_i_have_a_complete_referral` method. 

I wonder if this would be clearer if we simply created an incomplete referral here, and made all the record changes to "complete" the referral below (rather than rely on the `:complete` trait).

malcolmbaig (approved):

PR #234, opened 2022-11-28
Employer form - limit file upload types and size

steventux:
### Context

We should limit what users can upload as evidence and allegation files.

Adds a file upload validator - borrowed from Apply for Qualified Teaching Status and amended so that the validator can accept multiple files.

image

### Link to Trello card

https://trello.com/c/bOcayQul/995-rsm-employer-form-limit-file-upload-types-and-size

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


steventux:
Thanks @felixclack have amended the last commit as per your comments.

PR #237, opened 2022-11-28
Add a lint check for RSpec

felixclack:
There are variations in how we write our specs, which will likely
continue to vary as other people contribute to the project.

To reduce the cognitive load on the author when writing and the
maintainer when reading, we can auto-format our specs using the RSpec
cops for Rubocop.

These cops mostly follow the style described in `https://betterspecs.org`.

The exceptions added to `rubocop.yml` are to ensure the cops match the
testing styleguide we already have written.

Feel free to modify the cops if you have a strong opinion on any of
them.

This change brings in the default cops and ensures our codebase doesn't
have any violations.

Also, while there I noticed we have 2 time helpers included in the
project. Seeing as Rails provides this already, it makes sense to ditch
Timecop.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -2,6 +2,7 @@ inherit_gem:
   rubocop-govuk:
     - config/default.yml
     - config/rails.yml
+    - config/rspec.yml
 
 inherit_from:
   - node_modules/@prettier/plugin-ruby/rubocop.yml
@@ -16,6 +17,14 @@ Rails/ApplicationController:
 
 Rails/SaveBang:
   Enabled: false
+
+RSpec/Capybara/FeatureMethods:
+  Exclude:
+    - spec/system/**/*
+
+RSpec/InstanceVariable:
+  Exclude:
+    - spec/system/**/*
 
 Style/NumericLiterals:
   Exclude:
--- a/Gemfile
+++ b/Gemfile
@@ -34,7 +34,6 @@ group :development, :test do
   gem "pry"
   gem "pry-nav"
   gem "rspec-rails"
-  gem "timecop"
 end
 
 group :development do
@@ -51,9 +50,6 @@ end
 group :test do
   gem "capybara"
   gem "cuprite"
-end
-
-group :test do
   gem "rspec"
   gem "shoulda-matchers"
 end
--- a/spec/forms/users/otp_form_spec.rb
+++ b/spec/forms/users/otp_form_spec.rb
@@ -4,33 +4,30 @@ require "rails_helper"
 
 RSpec.describe Users::OtpForm do
   describe "#otp_expired?" do
-    it "is true if user otp is expired" do
-      Timecop.freeze do
-        user =
-          create(:user, otp_created_at: Users::OtpForm::EXPIRY_IN_MINUTES.ago)
-        form = described_class.new(id: user.id)
+    subject { form.otp_expired? }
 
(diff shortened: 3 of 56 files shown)


Gemfile
```diff
@@ -34,7 +34,7 @@ group :development, :test do
   gem "pry"
   gem "pry-nav"
   gem "rspec-rails"
-  gem "timecop"
+  gem "rubocop-rspec", require: false
```
vassyz, line 37: It's already a dependancy of `rubocop-govuk`. Check the current rules here: https://github.com/alphagov/rubocop-govuk/blob/ce655779e45892db6ac00d6789c652ec2e506748/config/rspec.yml `RSpec/ExampleLength`is already `false`

spec/forms/users/otp_form_spec.rb
```diff
@@ -4,33 +4,31 @@
...
-        user =
-          create(:user, otp_created_at: Users::OtpForm::EXPIRY_IN_MINUTES.ago)
-        form = described_class.new(id: user.id)
-
-        expect(form.otp_expired?).to eq true
-      end
+    subject { form.otp_expired? }
+
+    let(:form) { described_class.new(id:) }
+    let(:id) { user.id }
```
malcolmbaig, line 10: Doesn't look like we need this to be a separate let (not used anywhere on its own)

malcolmbaig (approved):
Linting the specs is a positive, and by and large the changes here are improvements in clarity. 

Happy to approve, but I'd encourage us not to be overly prescriptive when it comes to testing style. For example — with longer, more complex specs, leaning heavily on `let`s and a `subject` defined very early on in the file worsens readability, IMO. Some amount of setup local to the example block and locally-defined helper methods are valid techniques to deal with this. Leaving room for exercising that judgement would be helpful long term.

Would like to see another approval before merge given the wide range of the PR, but otherwise 👍🏼

app/forms/referrals/personal_details/age_form.rb
```diff
@@ -5,7 +5,7 @@ class AgeForm
       include ActiveModel::Model
       include ValidatedDate
 
-      attr_accessor :referral
+      attr_accessor :date_params, :referral
```
steventux, line 8: This and other date field changes don't seem to relate to the commit message. 
FWIW I think this is a worthwhile refactor but should either be in another PR or at least another commit, it doesn't relate to RSpec linting.

.rubocop.yml
```diff
@@ -10,13 +11,24 @@ inherit_mode:
   merge:
     - Exclude
 
+require:
```
vassyz, line 14: Is this needed?

steventux (changes requested):
It feels like there are a lot of changes in this PR which don't relate to RSpec linting so much as general refactoring, it would be good to pick these out either into a separate PR for review or at least separate commits for better future understanding.
eg. there's a refactor of how various date params are passed to forms which seems additional to RSpec linting.

app/forms/referrals/personal_details/age_form.rb
vassyz: Just wanted to say the same thing when I saw your comment. I was wondering if it could come out of a lint error, but don't think that is the case.

vassyz (approved):
Looks good. I think this PR does too much, but you might as well merge it before you have more conflicts to fix.

steventux (approved):

PR #244, opened 2022-11-30
Refactor #save to be consistent with Rails conventions

felixclack:
Passing params to the save method of a form object is different from the
Rails convention. This unexpected behaviour can make it harder for
someone reading the code in the future to understand quickly.

Moving the date_params to the constructor seemed like the simplest
change to make here. It ensures all params get instantiated in the same
way and lets the `save` method behave the same way as other Rails
models.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


vassyz (approved):
Looks great 👍

PR #245, opened 2022-11-30
Add test users to support interface

malcolmbaig:
### Context
Currently you can sign in to the service fairly easily on test environments by submitting your email address and copying the one-time password that's rendered just above the OTP form. This was a temporary solution that allowed sign up in the absence of any email functionality.

We shouldn't keep this around as it's a pretty big security hole if there's ever any kind of bug in the environment check or other related code.

Although we now have emails, it would be convenient to log in as a test user from the support console if anyone wants to test functionality.

- Add a section to support that allows us to create and sign in a user if we're on a test environment
- The same screen optionally allows sign in with an existing user account
- Add the `pagy` gem for pagination support

Screenshot_20221130_122031

### Guidance to review

- Fairly straightforward to test manually by navigating to the support interface
- Is the approach sound? Is there anything we might want to restrict/add at this stage?

### Link to Trello card

https://trello.com/c/A2qjN7q5
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


app/controllers/support_interface/test_users_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+    before_action :redirect_unless_test_environment
+
+    def index
+      @pagy, @users = pagy(User.newest_first)
+    end
+
+    def create
+      user = FactoryBot.create(:user)
+      sign_in(user)
+      redirect_to root_path
```
steventux, line 13: This threw me when testing locally, I think we need at least a flash message to say the user was created, I expected to be taken back to the test users index so I could see the newly created test user (and then sign in as that user myself)

steventux (approved):
This works well, although I found the test user creation -> sign in -> redirect to start of screener jump a bit misleading.

app/controllers/support_interface/test_users_controller.rb
malcolmbaig: Yeah good shout. I'll add both the flash and adjusted redirect.

PR #248, opened 2022-12-01
Handle stale otp form submissions

malcolmbaig:
### Context

If a user manually navigates back to the OTP screen after failed
authentication (eg - they've exhausted their OTP guesses), a stale form
submission gives them an error but doesn't redirect them anywhere
regardless of how many times they submit the form. This is because the
secret key has been wiped on their account, but there isn't any special
handling for that case.
Deal with this by sending the user down the expired OTP journey.

Also introduce a small semantic change to how OtpForm#otp_expired? works (see commits).

### Guidance to review

Steps to test:
- Be signed out
- Sign in
- Submit email
- Exhaust otp guesses (submit empty values)
- When redirected to the retry screen, hit back
- Submit the form
- Corrected behaviour should display the expired retry screen rather than allow infinite form submissions

### Link to Trello card
https://trello.com/c/pnaS8YMg

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/users/otp_controller.rb
+++ b/app/controllers/users/otp_controller.rb
@@ -19,7 +19,7 @@ class Users::OtpController < DeviseController
   def create
     @otp_form = Users::OtpForm.new(user_params)
 
-    if @otp_form.otp_expired?
+    if @otp_form.otp_expired? || !@otp_form.secret_key?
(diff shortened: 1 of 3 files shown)


steventux (approved):
:ship:

PR #249, opened 2022-12-01
Move date validation to DateValidator class

steventux:
### Context

We imported a lot of date logic from find-a-lost-trn as part of a form but have since used this in several places as a shared concern (thanks to @vassyz for shared examples and extracting the validation logic).

https://trello.com/c/dbcFXy8F/1006-refactor-date-validation
This PR moves the date validation into a validator class, the benefit being a slightly neater setup via the `validates` class method. 

- Adds `DateValidator` containing similar logic to the previous forms concern code.
- Converts existing forms with date validation to use the new validator.
- Removes date validating form concern.

### Link to Trello card

https://trello.com/c/dbcFXy8F/1006-refactor-date-validation

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


spec/forms/referrals/teacher_role/start_date_form_spec.rb
```diff
@@ -8,7 +8,7 @@
 
   let(:date_params) { nil }
   let(:referral) { build(:referral) }
-  let(:role_start_date_known) { true }
+  let(:role_start_date_known) { false }
```
vassyz, line 11: This change doesn't do anything. If can either be true or false. It's only used so that the default subject has a value on creation `described_class.new(date_params:, referral:, role_start_date_known:)`

vassyz (commented):
Should this still be a draft?

spec/forms/referrals/teacher_role/start_date_form_spec.rb
steventux: I think because the validator is now called with `valid?` (and not as part of `save`) it does make a difference now because the validator is configured like:

```ruby
validates :role_start_date, date: true, if: -> { role_start_date_known }
```

vassyz: Oh sorry, missed that.

vassyz (approved):

PR #255, opened 2022-12-02
Use uuid to identify users during sign in process

malcolmbaig:
### Context

When navigating to the OTP screen, we currently use the database id to identify the user signing in. Rather than expose this, we can switch to a UUID instead.

Add a uuid column to the users table, with uuid generation as a default value.

Use this identifier instead of the `id`.

### Link to Trello card

https://trello.com/c/a3lzNJYu
### Checklist

- [x] Attach to Trello card
- [ ] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/db/migrate/20221201155324_add_uuid_to_users.rb
+++ b/db/migrate/20221201155324_add_uuid_to_users.rb
@@ -1,0 +1,9 @@
+class AddUuidToUsers < ActiveRecord::Migration[7.0]
+  def change
+    enable_extension "pgcrypto" unless extension_enabled?("pgcrypto")
+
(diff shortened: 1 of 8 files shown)


db/migrate/20221201155324_add_uuid_to_users.rb
```diff
@@ -0,0 +1,9 @@
+class AddUuidToUsers < ActiveRecord::Migration[7.0]
+  def change
+    enable_extension "pgcrypto" unless extension_enabled?("pgcrypto")
```
vassyz, line 3: I prefer having a separate migration for enabling extensions.

vassyz (approved):

vassyz (approved):

PR #275, opened 2022-12-07
Append evidence uploads

steventux:
### Context

The evidence _check answers_ step should contain a button to allow the user to add more evidence.

- Allows user to add more evidence to a referral
- Validates that appended evidence doesn't exceed file limit

image

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


app/validators/file_upload_validator.rb
```diff
@@ -30,6 +30,13 @@ def validate_each(record, attribute, uploaded_files)
       record.errors.add(attribute, :file_count, max_files: MAX_FILES)
     end
 
+    if options[:append] && options[:existing_size_method]
```
vassyz, line 33: Why do you need `append`? Can't you just check `if options[:existing_size_method]`?

steventux: Yes good point :+1: will amend

app/validators/file_upload_validator.rb
```diff
@@ -30,6 +30,13 @@ def validate_each(record, attribute, uploaded_files)
       record.errors.add(attribute, :file_count, max_files: MAX_FILES)
     end
 
+    if options[:existing_size_method]
```
vassyz, line 33: I'm not entirely sure this is the best way of doing it. Feels like we should validate the number of evidence child records a referral can have.

steventux:
@vassyz I think there's a bigger issue with the `return_to` url here as adding more evidence means we have to categorise that evidence. For now this means running through all evidence categories steps, which isn't ideal. Trying to think of a better solution.

app/validators/file_upload_validator.rb
steventux: I've moved this validation onto the form and kept the validator as it was.

app/forms/referrals/evidence/categories_form.rb
```diff
@@ -44,24 +44,22 @@ def categories_other
       end
 
       def evidences
```
vassyz, line 46: Shouldn't we call this `uncategorised_evidences`?

steventux: I think we can do away with this method entirely now

app/forms/referrals/evidence/categories_form.rb
```diff
@@ -43,24 +43,18 @@ def categories_other
...
-        @evidences ||= referral.evidences
+      def next_evidence
+        referral.evidences.uncategorised.first
       end
 
-      def adjacent_evidence(next_item: true)
-        index = evidences.index(evidence)
+      def previous_evidence
+        index = referral.evidences.index(evidence)
         return if index.nil?
```
vassyz, line 52: Maybe I'm missing something, but wouldn't this approach `return if index.in? [0, nil]` work as well?

app/forms/referrals/evidence/upload_form.rb
```diff
@@ -29,7 +31,12 @@ def save
             )
           end
 
-        referral.update(evidences: evidences.sort_by(&:filename))
+        if referral.evidences.size + evidences.size > MAX_FILES
+          errors.add(:evidence_uploads, :file_count, max_files: MAX_FILES)
+          return false
+        end
+
+        referral.evidences << evidences.sort_by(&:filename)
```
vassyz, line 39: I know the sorting was already there, but why sort by filename? Wouldn't it make more sense to show the latest uploaded file at the bottom?

steventux: It's not really part of this PR and the existing functionality has been design reviewed but the rationale was that because the user can upload several files at once and we persist these records in one operation, timestamps might not produce an intuitive order or a predictable one.
We do rely a bit on order for the `next` step in categorisation.

spec/forms/referrals/evidence/upload_form_spec.rb
```diff
@@ -8,15 +8,7 @@
...
-      [
-        "",
-        Rack::Test::UploadedFile.new(
-          Rails.root.join("test/fixtures/files/doc2.pdf")
-        ),
-        Rack::Test::UploadedFile.new(
-          Rails.root.join("test/fixtures/files/doc1.pdf")
-        )
-      ]
+      ["", fixture_file_upload("doc2.pdf"), fixture_file_upload("doc1.pdf")]
```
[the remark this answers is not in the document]
steventux: There were not introduced in this PR.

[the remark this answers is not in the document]
vassyz: Ok, maybe we can have another PR to remove two of the files.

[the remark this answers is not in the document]
gpeng: That would be good. We should look at that as a tech debt thing separately so we can focus here 👍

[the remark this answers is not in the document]
steventux: @vassyz fwiw we assert differing upload file names in the system spec when adding and appending, perhaps it would be better to add a convention for the file names eg. `upload1... upload3`?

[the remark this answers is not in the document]
vassyz: @steventux I agree. I added `file.pdf` unaware that I will be testing a second file in the future. Feel free to rename/delete based on that naming convention.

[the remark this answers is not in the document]
steventux: @vassyz https://github.com/DFE-Digital/refer-serious-misconduct/pull/296

gpeng (approved):

malcolmbaig (approved):
Nice, tested locally and works as described. Think I may have found a bug while using the browser back button, but that's tangential to this PR. Will write a card after some more testing.

PR #278, opened 2022-12-08
Associate EligibilityCheck records with Referrals

malcolmbaig:
### Context

There's currently no way of identifying whether a referral is from an employer or a member of the public. We need to distinguish between the two to provide different journeys.

- Add a 1-1 association between an EligibilityCheck record and the Referral that gets created once the check is complete.
- Add convenience methods on Referral to query the type

### Guidance to review

~Raised as draft initially to discuss approach, specs to follow.~
~Is this approach sound?~

Aside from code review, a manual test could consist of going through the screener and observing the type of referral created in the console.

### Link to Trello card
https://trello.com/c/GgpmUmCx

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/referrals_controller.rb
+++ b/app/controllers/referrals_controller.rb
@@ -1,12 +1,17 @@
 class ReferralsController < Referrals::BaseController
   before_action :check_employer_form_feature_flag_enabled
   before_action :redirect_to_referral_if_exists, only: %i[new create]
+  before_action :redirect_to_screener_if_no_id_in_session, only: %i[new create]
 
   def new
   end
 
   def create
-    referral = current_user.referrals.create
+    eligibility_check =
+      EligibilityCheck.find_by!(id: session.delete(:eligibility_check_id))
+    referral =
+      current_user.referrals.create!(eligibility_check_id: eligibility_check.id)
+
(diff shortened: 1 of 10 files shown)


app/controllers/referrals_controller.rb
```diff
@@ -6,7 +6,12 @@ def new
   end
 
   def create
-    referral = current_user.referrals.create
+    eligibility_check =
+      EligibilityCheck.find_by(id: session[:eligibility_check_id])
```
vassyz, line 10: You can make this shorter if you want: `EligibilityCheck.find_by(id: session.delete(:eligibility_check_id))`

app/models/eligibility_check.rb
```diff
@@ -27,6 +29,10 @@ def reporting_as_employer?
     reporting_as&.to_sym == :employer
   end
 
+  def reporting_as_member_of_public?
```
vassyz, line 32: An alternative would be to have an `enum` and use the convenience methods that come with that.

steventux (approved):
:+1: Was able to step through screener and inspect the Referral / EligibilityCheck association in the console.

PR #279, opened 2022-12-08
Direct user to start of the evidence flow if not started yet

steventux:
### Context

The user should only be taken through the evidence flow if they've not started it before.

If the user has previously been through the evidence flow take them to the check-answers step.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/forms/referral_form.rb
+++ b/app/forms/referral_form.rb
@@ -90,7 +90,11 @@ class ReferralForm
         ),
         ReferralSectionItem.new(
           I18n.t("referral_form.evidence_and_supporting_information"),
-          referrals_edit_evidence_start_path(referral),
+          section_path(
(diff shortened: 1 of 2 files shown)


app/forms/referral_form.rb
```diff
@@ -1,6 +1,7 @@
 class ReferralForm
   include Rails.application.routes.url_helpers
   include ActiveModel::Model
+  include ReferralHelper
```
malcolmbaig, line 4: Is this used?

malcolmbaig (approved):
Works as described 👍🏼

app/forms/referral_form.rb
steventux: No - will remove

PR #287, opened 2022-12-12
Spike: simplify route declarations for employer form

malcolmbaig:
### Context

We have a fairly well established pattern for how our routes/controllers are structured for the employer form. This is probably a good time to refactor the route declarations so they're more condensed and easier to scan. 

This also potentially has implications for how we structure the public form work. If we go down the path of entirely separate routes/controllers for the second form, it'll be important to keep the `routes.rb` manageable.

Use other routing methods to define nearly the same resources.

For each section:
- Controller names are unchanged
- Path helpers for edit/update change slightly
- The URL for the edit page has `/edit` appended

I've applied this change to a subset of routes to illustrate the idea. If the consensus is 👍🏼 then I'll apply to the remaining ~140 lines worth of referral routes.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/components/personal_details_component.rb
+++ b/app/components/personal_details_component.rb
@@ -29,7 +29,7 @@ class PersonalDetailsComponent < ViewComponent::Base
           {
             text: "Change",
             href:
-              referrals_edit_personal_details_age_path(
+              edit_referral_personal_details_age_path(
                 referral,
                 return_to: request.url
               ),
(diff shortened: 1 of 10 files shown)


steventux:
Agree this is cleaner, good we're retaining the section level directory structure.

malcolmbaig:
Will raise a new PR when this is ready to review

PR #297, opened 2022-12-13
Public form / personal details name form amendments

steventux:
### Context

There's a shorter flow for members of the public filling out personal details for referrals.

- Adds a `public` namespace to routes
- Adds a helper method to determine paths for a given subsection
- Adds controllers/views specific to the public flow
- Brings name for design up to date with prototype
- Tests both flows with system specs

### Link to Trello card

https://trello.com/c/a3IMdNNl/1037-rsm-public-form-personal-details-name-form-amendments

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


vassyz (approved):

app/controllers/referrals/personal_details/name_controller.rb
```diff
@@ -34,7 +34,11 @@ def name_params
       end
 
       def next_path
-        referrals_edit_personal_details_age_path(current_referral)
+        if current_referral.from_employer?
```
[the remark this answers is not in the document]
steventux: @vassyz it's a fair point, just discussing with @malcolmbaig how we avoid such issues. I'll amend this as I think we'll use a separate controller / view

spec/system/referrals/public/user_adds_personal_details_spec.rb
```diff
@@ -0,0 +1,96 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+RSpec.feature "Personal details", type: :system do
+  include CommonSteps
+  include SummaryListHelpers
+
+  scenario "A memboer of the public adds personal details to a referral" do
```
malcolmbaig, line 8:
```suggestion
  scenario "A member of the public adds personal details to a referral" do
```

app/components/personal_details_component.rb
```diff
@@ -10,9 +11,11 @@ def rows
           {
             text: "Change",
             href:
-              edit_referral_personal_details_name_path(
+              subsection_path(
```
vassyz, line 14: I'm not sure if `subsection` is the best way of describing this. Is it because `public` is a subsection of referrals?

vassyz: Ok, so `personal_details_name` is the subsection. Maybe named arguments would help with that. I had one for `return_url`. It helps me figure out what arguments are used for. E.g. `subsection_path: :personal_details_name`.

steventux: Yeah I understand your concerns, I think named params would help here, I felt this was a bit messy, other ideas were:
`referral: section:, subection, action: nil, return_to: nil` as an args list.
I think I started with subsection_path as a method name because ReferralForm implemented `section_path` but then I renamed this anyway :D

config/routes.rb
```diff
@@ -172,6 +172,13 @@
       resource :declaration, only: %i[show create], controller: :declaration
       resource :confirmation, only: %i[show], controller: :confirmation
       resource :review, only: %i[show], controller: :review
+
+      namespace :public do
+        namespace :personal_details, path: "personal-details" do
+          resource :name, only: %i[edit update], controller: :name
+          resource :check_answers, path: "check-answers", only: %i[edit update]
+        end
+      end
```
malcolmbaig, line 181: Possible alternative — we could flatten some of the path and controller nesting by opting for something like

`public-referrals/4/edit`
`PublicReferrals::PersonalDetailsController`

Route declaration would be enclosed by:

```
  resources :public_referrals, except: %i[index], path: "public-referrals" do
    scope module: "public_referrals" do
    end
  end

```

app/controllers/referrals/public/personal_details/name_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+          )
+        end
+
+        def update_path
+          referral_public_personal_details_name_path(current_referral)
+        end
+      end
+    end
+  end
+end
```
malcolmbaig, line 17: This inheritance approach looks like it could work quite nicely. Worth running with for a few sections to see how it shapes up.

config/routes.rb
steventux: I like this - was beginning to feel like it was modules all the way down

spec/system/public-referrals/user_adds_personal_details_spec.rb
```diff
@@ -0,0 +1,96 @@
+# frozen_string_literal: true
```
malcolmbaig, line 1: nit: `public_referrals` probably more consistent for this folder name

malcolmbaig (approved):
LG 👍🏼

PR #301, opened 2022-12-14
Refactor route declarations

malcolmbaig:
### Context

We've agreed to refactor our route declarations to make them a bit simpler.

### Changes proposed in this pull request
Syntax changes as outlined in the original [spike](https://github.com/DFE-Digital/refer-serious-misconduct/pull/287).

### Guidance to review
Splitting this work into 2-3 PRs to keep the changes manageable. The next PR probably won't be as granular in its commits either.

### Link to Trello card
https://trello.com/c/e1imnfvC

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history
- [x] Tested by running locally

PR #302, opened 2022-12-14
Refactor route declarations - remaining edit actions

malcolmbaig:
### Context

Continuation of https://github.com/DFE-Digital/refer-serious-misconduct/pull/301

See commits

### Guidance to review
Part 2 of 3 — clearing away the edit actions and some of the more complex evidence routes. Will cover update paths in the next PR.

### Link to Trello card
https://trello.com/c/e1imnfvC

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/controllers/referrals/evidence/categories_controller.rb
+++ b/app/controllers/referrals/evidence/categories_controller.rb
@@ -16,12 +16,12 @@ module Referrals
           set_back_link
           next_evidence = @evidence_categories_form.next_evidence
           if next_evidence.present?
-            redirect_to referrals_update_evidence_categories_path(
+            redirect_to edit_referral_evidence_categories_path(
                           current_referral,
                           next_evidence
                         )
           else
-            redirect_to referrals_edit_evidence_check_answers_path(
+            redirect_to edit_referral_evidence_check_answers_path(
                           current_referral
                         )
           end
@@ -45,7 +45,10 @@ module Referrals
       helper_method :evidence
 
       def set_back_link
-        session[:evidence_back_link] = request.url
+        session[:evidence_back_link] = edit_referral_evidence_categories_path(
+          current_referral,
+          evidence
+        )
       end
     end
   end
--- a/app/components/evidence_component.rb
+++ b/app/components/evidence_component.rb
@@ -11,7 +11,7 @@ class EvidenceComponent < ViewComponent::Base
         actions: [
           {
             text: "Change",
-            href: referrals_edit_evidence_start_path(referral),
+            href: edit_referral_evidence_start_path(referral),
             visually_hidden_text: "evidence"
           }
         ],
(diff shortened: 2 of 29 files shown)


app/controllers/referrals/evidence/categories_controller.rb
```diff
@@ -45,7 +45,10 @@ def evidence
       helper_method :evidence
 
       def set_back_link
-        session[:evidence_back_link] = request.url
+        session[:evidence_back_link] = edit_referral_evidence_categories_path(
+          current_referral,
+          evidence
+        )
```
malcolmbaig, line 51: @steventux Flagging this for review in particular as just want to double check I haven't introduced a behaviour change here.

steventux: I think this is fine, it should always save the path of the current step, which is always the edit categories step, I may have originally thought it could vary.

steventux (approved):

app/controllers/referrals/evidence/categories_controller.rb
vassyz: This works fine if you only upload 2 files max. If you upload 3 then the 2nd `< Back` click will take you to the upload page, because you're only setting it in the `update` method. In the `edit` method you'd have to set it as some sort of `previous_evidence`.

steventux: @vassyz I think this was an issue I saw when adding the _Add more evidence_ functionality. 
_Previous_ evidence is meaningless in the context of the steps taken when appending evidence, because we denote appended evidence by a lack of categories. We'd have no real back route based on this criteria as the categories will have been saved. This is why I used `request.url` I think, is there a reason we shouldn't retain this?

vassyz: Yeah, then this doesn't work properly anyway. When you go back to the uploads page the ones you categorised are already gone, so there's so decent way of going back and forth between these pages.

steventux: It seems like we could also save the evidence back link from the upload page as it's an anomalous journey in terms of going back. Either way it evidently needs special attention and shouldn't block this PR.

malcolmbaig: @steventux Main reason for the change from `request.url` is that edit/update paths no longer have the same shape (they were previously identical, distinguished by the HTTP verb used, now edit paths have `/edit` appended).

vassyz: Can't we hide the button using the `evidence_back_link` for now?

malcolmbaig: Can we get this carded up to investigate and fix separately? Will merge in these route changes for now to minimise rebase issues.

steventux: @malcolmbaig Willdo :+1:

PR #310, opened 2022-12-16
Remove unused Referrals#show action

malcolmbaig:
### Context

When implementing part of the public form, I noticed that Referrals#show and #edit are the same view. I think we can get rid of #show.
Remove the show action and related view template. Update the few instances where it was being used (user sign in flow).

### Guidance to review
Looking at the git history, I think the show action was intended to be the review screen at some point. But we've since introduced the Review controller which takes on this responsibility.

Is my assessment correct here and can we ditch this action?

### Link to Trello card
na

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


malcolmbaig:
@felixclack would appreciate a review of this as I think you worked on related screens.

PR #315, opened 2022-12-19
Update "About their role" section

vassyz:
### Context

The "About their role" journey got updated.

* Reorder pages
* Add pages for end date, organisation address, reason leaving and new working location

### Guidance to review

* I have not addressed the back button issues.
* Accessing pages from the check answer page might not update the flow properly.
* Teaching somewhere else has been renamed to working somewhere else, but I have not updated the files and classes.
* The teacher role system spec doesn't test the content on the check answers page anymore because there are more variations of getting to that page and displaying content differently.

I will address these issues in future PRs.

### Link to Trello card

[](https://trello.com/c/iJKIsqQj/1036-update-about-their-role-section)

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/components/their_role_component.rb
+++ b/app/components/their_role_component.rb
@@ -7,159 +7,375 @@ class TheirRoleComponent < ViewComponent::Base
 
   def rows
     @rows = [
-      {
-        actions: [
-          {
-            text: "Change",
-            href:
-              edit_referral_teacher_role_start_date_path(
-                referral,
-                return_to: request.url
-              ),
-            visually_hidden_text: "start date"
-          }
-        ],
-        key: {
-          text: "Job start date"
-        },
-        value: {
-          text:
-            (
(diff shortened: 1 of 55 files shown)


steventux (approved):
I've tested locally, it all appears to work from the journeys I've attempted, changes to my initial answers, validations and section completion all appeared to work.

I think it's right to keep the system spec linear - they shouldn't test every possible combination of answers and user flow. 

Ideally this PR would contain more atomic commits, 1 monster commit makes reviewing harder and doesn't really add anything to commit history.

Obviously this was a very big refactor and it appears to work as expected, quite hard to feel like I've tested everything but the code looks good.

vassyz:
[Update "About their role" section](https://trello.com/c/iJKIsqQj/1036-update-about-their-role-section)

PR #320, opened 2022-12-21
Consider preprod a test environment

malcolmbaig:
Smoke test is currently failing because we switched on the referral form feature in preprod. We expect test environments to have this switched on, and production to have it switched off for now.

Add preprod to the list of test environments.


tvararu (approved):
Thanks Malcolm, I thought of this yesterday but forgot to raise it...

PR #321, opened 2022-12-22
Remove job title from the public flow

felixclack:
When someone is referring as a member of the public, we don't need to
collect their job title. It is not relevant.

I considered creating a new route and controller for this flow, as per
the approach in #297.

This seemed overkill as the only difference we have in this particular
case is to change the value of `next_path` for the name controller.

It seems like we're missing an orchestration object where we could
define the questions and their order in a more declarative fashion.

I opted not to do that here as it doesn't seem like the right time to do
it.

Instead, I went for the simplest change possible, which is to use a
conditional statement to determine the next question in the name
controller.

### Link to Trello card

[Trello](https://trello.com/c/FRqOOzVo/1044-s-m-public-form-only-remove-job-title-page)

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/spec/system/public_referrals/user_adds_your_details_spec.rb
+++ b/spec/system/public_referrals/user_adds_your_details_spec.rb
@@ -1,0 +1,55 @@
+require "rails_helper"
+
+RSpec.feature "Public flow: Your details", type: :system do
+  include CommonSteps
+  include SummaryListHelpers
+
+  scenario "adding your details" do
+    given_the_service_is_open
+    and_i_am_signed_in
+    and_the_referral_form_feature_is_active
+    and_i_am_a_member_of_the_public_with_an_existing_referral
+    and_i_visit_the_referral
+    and_i_click_on_the_your_details_task
+    then_i_see_the_your_name_page
+
+    when_i_enter_my_name
+    and_i_click_save_and_continue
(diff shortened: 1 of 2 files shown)


malcolmbaig:
I think we likely need to remain consistent with the approach outlined in https://github.com/DFE-Digital/refer-serious-misconduct/pull/297. Coupled with https://github.com/DFE-Digital/refer-serious-misconduct/pull/323, which builds on top of it, employer and public referrals are edited in two distinct namespaces. Moving between these namespaces to edit a public referral could get confusing.

spec/system/public_referrals/user_adds_your_details_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    given_the_service_is_open
+    and_i_am_signed_in
+    and_the_referral_form_feature_is_active
+    and_i_am_a_member_of_the_public_with_an_existing_referral
+    and_i_visit_the_referral
+    and_i_click_on_the_your_details_task
+    then_i_see_the_your_name_page
+
+    when_i_enter_my_name
+    and_i_click_save_and_continue
```
vassyz, line 17: I'd probably add one more `click_save_and_continue` to get to the check answers page.

PR #323, opened 2022-12-22
Add routing for public form task list

malcolmbaig:
### Context

Prior to adding the public form task list as a hub for the various sections, we need to ensure we're directing users down the correct routes depending on the type of referral they're completing.

Going through the screener results in either a public or employer
referral being created. This commit ensures that referrals are
handled by separate public or employer controllers. This structure is
being trialled so we can assess ways to deal with the complexity of two
distinct journeys that share many of the same steps.

Key changes:

- Add a PublicReferralsController that inherits from the existing
  ReferralsController
- Add a helper to pick the correct edit route for redirects
- Ensure we end up on the correct new referral page immediately after
  the screener

See full commit list for other, related changes.
### Guidance to review

The system specs outline the intended behaviour, but this can be tested locally by completing the screener with both employer and public referrals. These should lead to task lists sitting at  `/referrals/:id/edit` and `/public-referrals/:id/edit` paths, respectively.

### Link to Trello card
https://trello.com/c/SiOxJOXb

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/concerns/referral_paths.rb
+++ b/app/controllers/concerns/referral_paths.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+module ReferralPaths
+  extend ActiveSupport::Concern
+
+  def edit_path_for(referral)
+    if referral.from_member_of_public?
+      edit_public_referral_path(referral)
+    else
+      edit_referral_path(referral)
+    end
+  end
+end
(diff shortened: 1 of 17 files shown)


steventux (approved):
:+1: Looks good. 
It would be good to revisit the various path helpers and concerns at some point in the future to see if the logic can be merged.
Thanks for fixing the `public_referrals` path naming issue.

PR #324, opened 2022-12-22
Employer form evidence flow amendments

steventux:
### Context

We have iterated the design of the evidence flow for both public and employer forms. We no longer categorise evidence and the flow to add more evidence is simplified.

- Remove evidence categorisation
- Add a new uploaded evidence list with associated actions
- Amend employer evidence flow content
- Amend evidence flow back links
- Add conditional redirect when deleting evidence

### Guidance to review

- Public evidence flow amendments to follow
- The `uploaded` action is dissimilar to other actions as it only routes the user to one of 3 places, it doesn't persist anything. I didn't give this too much thought at the time but suggestions welcome on how to improve this.

### Link to Trello card

https://trello.com/c/ytCuCtmi/1023-public-employer-form-evidence-check-answers-add-content-links
https://trello.com/c/XJDfaYrI/1027-public-employer-form-add-more-evidence

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


app/controllers/referrals/evidence/uploaded_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+    class UploadedController < Referrals::BaseController
+      def edit
+        case params[:more_evidence]
+        when "yes"
+          redirect_to edit_referral_evidence_upload_path(current_referral)
+        when "no"
+          redirect_to edit_referral_evidence_check_answers_path(
+                        current_referral
+                      )
+        else
```
vassyz, line 12: Usually how we handle this is with an error similar to `Select if you have evidence to upload`.

db/schema.rb
```diff
@@ -132,8 +131,8 @@
     t.string "duties_details"
```
vassyz, line 131: I try not to add the schema changes that only shuffle fields around. A solution to this would be to reset the schema before creating a migration so you keep the db structure from he `main` branch.

app/components/evidence_component.rb
```diff
@@ -4,68 +4,62 @@ class EvidenceComponent < ViewComponent::Base
   attr_accessor :referral
 
   def rows
-    return evidence_rows if referral.evidences.any?
+    [anything_to_upload_row, evidence_row]
```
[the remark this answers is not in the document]
steventux: We probably shouldn't display this row if no evidence was uploaded - will amend

app/controllers/referrals/evidence/uploaded_controller.rb
steventux: This is admittedly a bit of an odd case, no data gets persisted here, it's just a routing switch determined by the `more_evidence` param or lack of it. 
I wondered about doing this instead of the usual pattern of `edit` + `update` but settled on this approach as we're not updating anything and having 2 actions seemed superfluous. 
Essentially the same action is called once without params to render the form/view and on form submission we're routed to either the upload or the check answers path. Maybe it would be clearer separared out?

app/views/referrals/evidence/uploaded/edit.html.erb
```diff
@@ -0,0 +1,38 @@
...
+              Delete<span class="govuk-visually-hidden"> <%= evidence.filename %></span>
+            <% end %>
+          </dd>
+        </div>
+        </li>
+      <% end %>
+    </dl>
+
+    <%= form_with url: edit_referral_evidence_uploaded_path(current_referral), method: :get do |f| %>
+      <%= f.govuk_radio_buttons_fieldset(:more_evidence, legend: { size: "m", text: "Do you have evidence to upload?" }) do %>
```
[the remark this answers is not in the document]
steventux: I've stuck to the design prototype here.

app/controllers/referrals/evidence/uploaded_controller.rb
vassyz: I think the code is clear enough. The issue is that if you submit without selecting it doesn't follow the pattern of showing a `Select if` error. An easy solution would be to show a flash message, but that breaks another pattern. The error should be shown on the form, and not in a generic message, but then you'd need to use `ActiveModel`, so I really don't know what the best option would be.

vassyz (approved):

db/schema.rb
steventux: Ah! I see - will follow up with a meaningful error in another PR.

vassyz: I tagged the wrong line number here by mistake.

app/controllers/referrals/evidence/uploaded_controller.rb
steventux: Ah! I understand, will follow up with a PR to fix this

steventux: @vassyz I've opened https://github.com/DFE-Digital/refer-serious-misconduct/pull/338 as a follow up on this

PR #341, opened 2022-12-30
Fix teacher role back links

vassyz:
### Context

The back links for the teacher journey were't pointing to the right pages. An exception (for now) is the start date page that always goes back to the same organisation page because otherwise it would have to dynamically set rules based on the same organisation answer.

### Link to Trello card

https://trello.com/c/SKTTPWCb/1050-fix-back-links-for-about-their-role-section

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


vassyz:
[Fix back links for "About their role" section](https://trello.com/c/SKTTPWCb/1050-fix-back-links-for-about-their-role-section)

gpeng (approved):

PR #343, opened 2022-12-30
Add a check for referral type in controllers

malcolmbaig:
### Context
When retrieving a referral from the database, it would be useful to have a check which ensures that the referral being used in any given controller is the correct type for the current namespace (ie - public vs employer referral).

This commit adds a controller concern which raises an error if the current path doesn't correspond to the referral being retrieved.

This ensures we don't accidentally pass users into the employer flow while building out the public one.

### Guidance to review
Was building out the public form hub page and this seemed like a thing we'd need as we progress with the current architecture of having two distinct controller namespaces for the public/employer forms. Interested in feedback on the approach, potential issues, alternatives, etc.

One side effect of this is that we probably need to complete sections of the public form in larger chunks than we're doing currently, or at least leave TODOs in our system specs that we follow up on in subsequent cards. Reason being that this code will cause the app to error out if the next step in a public form journey is an employer form route.

### Link to Trello card

https://trello.com/c/SiOxJOXb
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


app/controllers/concerns/ensure_referral_type_matches_path.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+module EnsureReferralTypeMatchesPath
+  extend ActiveSupport::Concern
+
+  class ReferralTypeDoesNotMatchPath < StandardError
+  end
+
+  def check_referral_against_path(referral)
+    if referral.from_employer? && !controller_path.match?(/^referrals/)
```
vassyz, line 9: This works, but I'm wondering if it won't cause issues in the future. You can tag the scope more explicitly like `scope module: :referrals, employer: "true" do` and check if `params[:employer]` is true.

app/controllers/referrals_controller.rb
```diff
@@ -1,5 +1,6 @@
 class ReferralsController < Referrals::BaseController
   include ReferralPaths
+  include EnsureReferralTypeMatchesPath
```
vassyz, line 3: Do you need this `include`?

app/controllers/referrals/base_controller.rb
```diff
@@ -15,6 +16,7 @@ def edit
     def current_referral
       id = params[:referral_id] || params[:public_referral_id]
       @current_referral ||= current_user.referrals.find(id)
+      @current_referral.tap { |r| check_referral_against_path(r) }
```
vassyz, line 19: Maybe I'm missing something, but I'm not sure the `tap` method is used optimally here (and [below](https://github.com/DFE-Digital/refer-serious-misconduct/pull/343/files#diff-1707f3893d1d3e8d39ba65e031387381f7c0def8b0dbd4d6681c8bb5037ea4fcR56)). Although the `find` method gets memoized, you will be calling `check_referral_against_path` every time you access `current_referral`. Wouldn't this be a better way?

```ruby
@current_referral ||=
  current_user
    .referrals
    .find(id)
    .tap { |r| check_referral_against_path(r) }
```

malcolmbaig: Yes, definitely. Updated 👍🏼

app/controllers/referrals_controller.rb
malcolmbaig: Good catch, removed.

app/controllers/concerns/ensure_referral_type_matches_path.rb
malcolmbaig: Not a bad idea, but it does leave us with a slightly redundant param hanging around in the URL. 

Open to other ideas on this. I suspect the above will blow up fairly reliably if we try to implement any functionality where the path and referral type don't line up. Particularly if we do the refactor to rename the `referral` namespace to `employer_referral`.

malcolmbaig: >I can't find the original discussion around why we chose the current approach

Spoke with @felixclack offline, summary:

- This is an approach we're trialling for the public form to manage the forking complexity of the two distinct form journeys. 
- We can review once a few sections are done and potentially spike an alternative.

A route constraint looks like a good idea. One issue is that it isn't straightforward to obtain the current user for scoping retrieval of the referral. This may mean we just use the constraint for checking the referral type, and leave referral querying to the controllers. Will update 👍🏼

malcolmbaig:
@felixclack @vassyz Have implemented a route constraints version of this [here](https://github.com/DFE-Digital/refer-serious-misconduct/pull/357). Seemed tidier than messing around with a rebase and adding more content to this PR.

PR #357, opened 2023-01-06
Add checks to ensure consistency between referral id and URL path

malcolmbaig:
### Context

An alternative solution to https://github.com/DFE-Digital/refer-serious-misconduct/pull/343/, based on feedback provided in that PR.

In brief: we're trialling an approach of using separate controller namespaces for the two kinds of referral to manage some of the complexity around implementing two similar flows that diverge in several places. To support this, it would be useful to have some code that ensures the correct referral type is being edited at a given URL path.

- Scope retrieval in the hub page controllers using scopes on the Referral model
- Use route constraints to make a similar check on all sub-resources
- Fixing some existing bugs that were flushed out by this change
- Deactivate the public form evidence system spec for now - the changes required here would bloat this PR, they'll be picked up subsequently (I've noted this on the [public form epic card](https://trello.com/c/XaAWBPcu)).

- I've also added config/routes.rb to the .prettierignore as I think it detracts from the readability of this file. I can drop this particular commit if this change is contentious at all.

### Guidance to review
- One manual test case would be to start with a public form and navigate to one of the steps that's been implemented (eg—Their details). Changing the URL from `public-referral` to `referral` should blow up.
- A similar thing should happen with a URL change on the hub page.

### Link to Trello card
Doing this as part of https://trello.com/c/SiOxJOXb, but only tangentially related to that card

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/lib/route_constraints/employer_constraint.rb
+++ b/app/lib/route_constraints/employer_constraint.rb
@@ -2,9 +2,7 @@ module RouteConstraints
   class EmployerConstraint
     def matches?(request)
       id = request.params[:referral_id]
-      referral = Referral.employer.where(id:).first
-
-      referral.present?
+      Referral.employer.exists?(id)
     end
   end
 end
--- a/app/lib/route_constraints/employer_constraint.rb
+++ b/app/lib/route_constraints/employer_constraint.rb
@@ -1,0 +1,10 @@
+module RouteConstraints
+  class EmployerConstraint
+    def matches?(request)
+      id = request.params[:referral_id]
+      referral = Referral.employer.where(id:).first
+
+      referral.present?
+    end
+  end
+end
--- a/.prettierignore
+++ b/.prettierignore
@@ -23,6 +23,7 @@ Procfile
 Procfile.*
 app/assets/builds
 bin/dev
+config/routes.rb
 data
 db/*.sqlite3
 db/schema.rb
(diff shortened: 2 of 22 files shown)


app/lib/route_constraints/employer_constraint.rb
```diff
@@ -0,0 +1,10 @@
+module RouteConstraints
+  class EmployerConstraint
+    def matches?(request)
+      id = request.params[:referral_id]
+      referral = Referral.employer.where(id:).first
+
+      referral.present?
```
vassyz, line 7: Can't this be replaced with `Referral.exists?(id)`?

app/lib/route_constraints/employer_constraint.rb
```diff
@@ -0,0 +1,10 @@
+module RouteConstraints
+  class EmployerConstraint
+    def matches?(request)
+      id = request.params[:referral_id]
+      referral = Referral.employer.where(id:).first
```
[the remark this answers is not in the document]
vassyz: I'll try to add this in a different PR.

app/lib/route_constraints/employer_constraint.rb
vassyz: I'm assuming it can. Done [here](https://github.com/DFE-Digital/refer-serious-misconduct/pull/357/commits/56af867b167f2a11df375d9ce6fd9d30dc827c24#diff-4fddc7d77aa492e591029d5c3a4f73f085232d87bb430b2645a1e789e604e1e8L7).

PR #369, opened 2023-01-13
Fix file deletions for public evidence

felixclack:
The public flow for adding evidence has an issue with deleting uploaded
files.

The links point to the employer flow, which leads to 404s.

This change enables the tests and ensures the paths are all correct.

### Link to Trello card

https://trello.com/c/HzpGmD6I/1095-fix-file-deletion-in-public-supporting-information-section

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

Screenshot 2023-01-13 at 10 50 23 am

### DIFF
--- a/app/components/evidence_component.rb
+++ b/app/components/evidence_component.rb
@@ -1,5 +1,6 @@
 class EvidenceComponent < ViewComponent::Base
   include ActiveModel::Model
+  include ReferralHelper
 
   attr_accessor :referral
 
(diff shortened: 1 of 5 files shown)


app/components/evidence_component.rb
```diff
@@ -15,7 +16,12 @@ def anything_to_upload_row
         {
           text: "Change",
           href:
-            edit_referral_evidence_start_path(referral, return_to: request.url)
+            subsection_path(
+              action: :edit,
+              referral:,
+              return_to: request.url,
```
gpeng, line 22: I think these might get pulled up in the pentest. Not so much here but when we render the links with them as we're not whitelisting them. Does `request.path` work here? We're going to have to change them all over the place anyway so not a biggy for now.

gpeng (approved):

PR #373, opened 2023-01-15
Fix validation errors for radio fields

vassyz:
Fixes the bug where clicking on validation errors would not jump to the invalid radio field.

### Link to Trello card

https://trello.com/c/BrtfsyCM/1104-fix-validation-errors-for-radio-fields-that-dont-go-to-the-field-when-clicked

### Checklist

- [x] Attach to Trello card
- [] Rebased main
- [x] Cleaned commit history


vassyz:
[Fix validation errors for radio fields that don't go to the field when clicked](https://trello.com/c/BrtfsyCM/1104-fix-validation-errors-for-radio-fields-that-dont-go-to-the-field-when-clicked)

PR #377, opened 2023-01-16
Deploy pentest environment

NickGraham101:
## Do Not Merge

This PR has been created to facilitate deployment of the pentest environment.  It is not intended to be merged into main.  Closing this PR will destroy the pentest environment once that is no longer required.  It won't destroy the DNS record, see https://trello.com/c/CoawQFko/1099-set-up-pentest-environment for instructions on that


gpeng:
This is looking good @NickGraham101 . I've had a click around and it all seems good although it seems very slow. Also we'll need a proper Notify key. I've created separate keys for each env so I can do another one for this.

gpeng:
Closing as we've finished with the pentest now.

PR #378, opened 2023-01-16
Ensure the check answers pages have back links 

felixclack:
- Replace subsection_path with polymorphic_path
- Ensure the check answers pages have back links

### Context

Not all the check answers pages have a back link and the ones that do
often link to the previous page.

The latest designs call for the back link on the check answers page to
always return to the main page for the referral.

While making this change, I noticed that the 'Your details' back links
only work in the context of an employer referral. There is a fix for
that here too.

### Link to Trello card

https://trello.com/c/MFB8bLU6/1090-defect-all-check-answers-pages-within-each-task-need-to-have-a-back-link-back-to-the-task-list

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/views/public_referrals/evidence/check_answers/edit.html.erb
+++ b/app/views/public_referrals/evidence/check_answers/edit.html.erb
@@ -1,5 +1,5 @@
 <% content_for :page_title, "#{'Error: ' if @evidence_check_answers_form.errors.any?}Check and confirm your answers" %>
-<% content_for :back_link_url, return_to_session_or(edit_public_referral_evidence_uploaded_path(current_referral)) %>
+<% content_for :back_link_url, return_to_session_or(polymorphic_path([:edit, current_referral.routing_scope, current_referral])) %>
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds-from-desktop">
--- a/app/controllers/public_referrals/personal_details/check_answers_controller.rb
+++ b/app/controllers/public_referrals/personal_details/check_answers_controller.rb
@@ -8,10 +8,6 @@ module PublicReferrals
       def update_path
         public_referral_personal_details_check_answers_path(current_referral)
       end
-
-      def back_link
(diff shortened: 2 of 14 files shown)


app/models/referral.rb
```diff
@@ -30,6 +30,12 @@ class Referral < ApplicationRecord
           )
         }
 
+  def routing_scope
+    return :public if from_member_of_public?
+
+    nil
```
vassyz, line 36: Is this necessary?

app/views/public_referrals/evidence/check_answers/edit.html.erb
```diff
@@ -1,5 +1,5 @@
 <% content_for :page_title, "#{'Error: ' if @evidence_check_answers_form.errors.any?}Check and confirm your answers" %>
-<% content_for :back_link_url, return_to_session_or(edit_public_referral_evidence_uploaded_path(current_referral)) %>
+<% content_for :back_link_url, return_to_session_or(polymorphic_path([:edit, current_referral.routing_scope, current_referral])) %>
```
vassyz, line 2: Worth moving `polymorphic_path([:edit, current_referral.routing_scope, current_referral])` to a helper?

app/models/referral.rb
vassyz, line 36: Is this needed?

vassyz (approved):

app/models/referral.rb
```diff
@@ -30,6 +30,12 @@ class Referral < ApplicationRecord
           )
         }
 
+  def routing_scope
```
gpeng, line 33: I'm not sure about this on the model as it doesn't seem like a model concern. Given we're pressed for time at the mo I'll card it up to think about as a refactor later. It's not a biggy. Should it return `:employer` rather than `nil` if it's not public?

gpeng (approved):

vassyz (approved):

PR #382, opened 2023-01-17
Use scoped routes where possible

felixclack:
Now that the referrals are scoped by type, we need to ensure links
take the scope into account.

The easiest way to do this is to use the Rails link helpers and pass the
routing scope. If it is `nil` then the default scope is used, eg.
`referrals`, alternatively we use `public_referrals`.

There are some views and controllers that are only used for the employer
flow, so I left some of those links as they are.

I attempted to change the uploaded evidence views but ran into an issue
with the routes not matching the model they refer to, eg.
`ReferralEvidence` is rendered at the `referrals/1/evidences/1` route.

The custom route means we need to override more things and I was trying
to limit the scope of this change to simply replacing the helper methods
with the array style.

### DIFF
--- a/app/components/contact_details_component.rb
+++ b/app/components/contact_details_component.rb
@@ -10,11 +10,7 @@ class ContactDetailsComponent < ViewComponent::Base
         actions: [
           {
             text: "Change",
-            href:
-              edit_referral_contact_details_email_path(
-                referral,
-                return_to: request.url
-              ),
+            href: path_for(:email),
             visually_hidden_text: "email"
           }
         ],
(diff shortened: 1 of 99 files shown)


vassyz:
There might be a reason, but just making sure you're aware you left a few `update_path`s in the code. The ones for the public/employer controllers for referrals, declaration and referrer phone. I think the one in `PublicReferrals::Allegation::DetailsController` can be removed as it doesn't seem to be used anywhere.

vassyz (approved):

PR #388, opened 2023-01-18
Display a simple referral page for case workers

felixclack:
As part of building the case worker management interface, we need a page
that displays the details of a specific referral.

To make it easier for us to work on this page concurrently, I'm shipping
a very basic version that we can iterate on.

### Link to Trello card

https://trello.com/c/HEb0DEie/1109-case-worker-referral-show-page

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

Screenshot 2023-01-18 at 1 01 40 pm

### DIFF
--- a/app/models/referral.rb
+++ b/app/models/referral.rb
@@ -30,6 +30,8 @@ class Referral < ApplicationRecord
           )
         }
 
+  delegate :name, to: :referrer, prefix: true, allow_nil: true
+
   def routing_scope
     return :public if from_member_of_public?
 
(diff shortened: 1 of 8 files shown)


app/models/referral.rb
```diff
@@ -30,6 +30,8 @@ class Referral < ApplicationRecord
           )
         }
 
+  delegate :name, to: :referrer, prefix: true, allow_nil: true
```
vassyz, line 33: Can you add this to the index view as well?

vassyz (approved):

PR #393, opened 2023-01-19
Add all sections to case worker referral view

felixclack:
This adds all the sections to the case worker referral view that are
detailed in the current designs.

There are some parts to this that aren't implemented yet.

The referrer name is stored as a single field. For now I opted to split
this value into first and last name methods so we can render the
proposed UI.

When the change to the way we collect this value happens, then these
methods will continue to work and we won't need to change the referral
view.

Secondly, we don't collect the referrer's current employer address.

Rather than guess at how we might do that, I left it unimplemented.

### Link to Trello card

https://trello.com/c/HEb0DEie/1109-case-worker-referral-show-page

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/helpers/referral_helper.rb
+++ b/app/helpers/referral_helper.rb
@@ -30,4 +30,39 @@ module ReferralHelper
       "Incomplete"
     end
   end
+
+  def allegation_details(referral)
+    if referral.allegation_upload.attached?
+      [
+        "File:",
+        govuk_link_to(
+          referral.allegation_upload.filename,
+          rails_blob_path(referral.allegation_upload, disposition: "attachment")
+        )
+      ].join(" ").html_safe
+    elsif referral.allegation_details.present?
+      referral.allegation_details.truncate(150)
+    else
+      "Incomplete"
+    end
+  end
+
+  def previous_allegation_details(referral)
+    if referral.previous_misconduct_upload.attached?
+      [
+        "File:",
+        govuk_link_to(
+          referral.previous_misconduct_upload.filename,
+          rails_blob_path(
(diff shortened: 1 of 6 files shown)


gpeng (approved):

app/models/referrer.rb
```diff
@@ -1,6 +1,14 @@
 class Referrer < ApplicationRecord
   belongs_to :referral
 
+  def first_name
+    name.split(" ").first
+  end
+
+  def last_name
```
vassyz, line 8: Won't this be the middle name in some cases? Why not last?

app/helpers/referral_helper.rb
```diff
@@ -30,4 +30,39 @@ def employment_status(referral)
       "Incomplete"
     end
   end
+
+  def allegation_details(referral)
+    if referral.allegation_upload.attached?
+      [
+        "File:",
```
vassyz, line 37: Does it need `"File:"`?

app/helpers/referral_helper.rb
```diff
@@ -30,4 +30,39 @@ def employment_status(referral)
...
+        govuk_link_to(
+          referral.previous_misconduct_upload.filename,
+          rails_blob_path(
+            referral.previous_misconduct_upload,
+            disposition: "attachment"
+          )
+        )
+      ].join(" ").html_safe
+    elsif referral.previous_misconduct_details.present?
+      simple_format(referral.previous_misconduct_details)
```
vassyz, line 63: One description truncates to 150 and the other uses `simple_format`. Shouldn't they be consistent?

PR #394, opened 2023-01-19
Add case worker zip file download

vassyz:
### Context

This is the first iteration of the download functionality for the case worker manage referrals page.

It adds a link to download the exported zip file:
Screenshot 2023-01-20 at 11 55 09

- select all the file attachments for the referral: `duties_upload`, `allegation_upload`, and all the documents attached via the submitted evidences.
- group the files into folders
- the downloaded file will have the following format for its name: `<timestamp>-referral-<referral_id>.zip`
- make sure the temp file gets deleted afterwards.

The way we handle this might change in the future as for larger files an asynchronous approach would be better (scheduled exports or exports triggered on click followed by an email with the download link.)

### Link to Trello card

https://trello.com/c/HontnuhK/1111-case-worker-zip-file-download

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/manage_interface/referrals_controller.rb
+++ b/app/controllers/manage_interface/referrals_controller.rb
@@ -1,12 +1,30 @@
 module ManageInterface
   class ReferralsController < ManageInterfaceController
+    before_action :referral, only: %i[show download]
+
     def index
       @referrals_count = Referral.submitted.count
       @pagy, @referrals = pagy(Referral.submitted)
     end
 
     def show
-      @referral = Referral.find(params[:id])
+    end
+
+    def download
+      filepath = @referral.zip_file_path
+      file = File.open(filepath)
+      contents = file.read
+      file.close
+
+      File.delete(filepath) if File.exist?(filepath)
+
+      send_data(contents, filename: @referral.zip_file_name)
+    end
+
+    private
+
+    def referral
+      @referral ||= Referral.find(params[:id])
     end
   end
 end
--- a/app/models/concerns/zippable.rb
+++ b/app/models/concerns/zippable.rb
@@ -1,0 +1,65 @@
+require "zip"
+
+module Zippable
+  extend ActiveSupport::Concern
+
+  included do
+    def zip_file_path
+      @zip_file_path ||= zip_file.path
+    end
+
+    def zip_file_name
+      "#{Time.zone.now.to_fs(:number)}-#{self.class.name.downcase}-#{id}.zip"
+    end
+
+    def has_attachments?
+      has_own_attachments? || has_evidence_attachments?
+    end
+
+    private
+
+    def zip_file
+      temp_file = Tempfile.new("temp.zip")
+
+      Zip::File.open(temp_file.path, create: true) do |zip_file|
+        attachments.each do |attachment|
+          next unless attachment.attached?
+
+          file = ActiveStorage::Blob.service.path_for(attachment.key)
+          zip_file.add(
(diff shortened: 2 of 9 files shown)


app/models/concerns/zippable.rb
```diff
@@ -0,0 +1,61 @@
...
+            File.join(file)
+          )
+        end
+      end
+
+      temp_file.path
+    end
+
+    def zip_file_name
+      "#{Time.zone.now.to_fs(:number)}-#{self.class.name.downcase}-#{id}.zip"
```
gpeng, line 26: I think we'll need to think about this with @adamsilver. There was 'resistance' to using our ID when we talked about this today but I think we'll need to make a case for 'an ID'. This is fine for now though.

gpeng (commented):
I think this looks good to get us going. We can probably attach the PDF as an attachment when we get to it and this will 'just work ™️'. We'll need to do some testing with a decent number of files as I suspect it might be pretty slow retrieving the files sequentially like this if there are a few or they're big but we can iterate on that.

app/models/concerns/zippable.rb
vassyz: You commented on the zip name. Yes, I can remove that id but wanted a reference to the referral.

gpeng: Don't remove it. We do need to reference the referral but we just need to have a think about referencing it in a way that's meaningful to the users. We unfortunately don't know the meaningful ID yet so we have a small content problem to solve here. Fortunately we have a crack team of experts for that (cc/ @Dan13L3, @adamsilver)

vassyz: Maybe we can generate a unique reference number when a referral gets created.

gpeng: We definitely can and sort of do with the database ID but there's a bit of concern that is confusing as the CMS generates an ID in the next step that everyone is used to using. We should definitely create the ID at the start and it be used throughout but there's a bit of tricky transition to manage.

gpeng: We're discussing this offline at the mo. Let's just go with this for now and we can update it when we have a decision about the referral ID.

vassyz:
[Case worker zip file download](https://trello.com/c/HontnuhK/1111-case-worker-zip-file-download)

app/controllers/manage_interface/referrals_controller.rb
```diff
@@ -1,12 +1,30 @@
...
     def index
       @referrals_count = Referral.submitted.count
       @pagy, @referrals = pagy(Referral.submitted)
     end
 
     def show
-      @referral = Referral.find(params[:id])
+    end
+
+    def download
```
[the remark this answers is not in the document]
vassyz: You mean having a separate downloads resource? I could, I guess it felt more like a referral to me, unless we persist the downloads. I was thinking of having it as a format for show referrals.

app/models/concerns/zippable.rb
```diff
@@ -0,0 +1,65 @@
+require "zip"
+
+module Zippable
```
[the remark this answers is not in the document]
vassyz: The idea was that it wouldn't get tied to referral. That it would be a concern that allows to download all attachments of a model. But then I realised that the evidence attachments are separate, so this became a lot less flexible than I'd have wanted. And we'll have to add the referral pdf export as well. I'll try to refactor.

gpeng:
Discussed offline. We'll address the refactor comments in a new PR.

gpeng (approved):

PR #412, opened 2023-01-25
Make error messages for date validations more accessible

paulodeon:
### Context

To improve the accessibility of the referral form error messages for the IAC audit

* Changes to the locale.yml
* Tweak to the date validator
Before

After
image
image
image

### Guidance to review

Go through the referral form and interact with the date fields attempting to save them in an incomplete state

### Link to Trello card

[](https://trello.com/c/3ZLNfEl2/1126-update-error-messages-for-date-validations)

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/validators/date_validator.rb
+++ b/app/validators/date_validator.rb
@@ -61,8 +61,13 @@ class DateValidator < ActiveModel::EachValidator
       return false
     end
 
-    if year < 1000
+    if year.zero?
       record.errors.add(attribute, :missing_year)
+      return false
+    end
+
+    if year < 1000
+      record.errors.add(attribute, :incomplete_year)
       return false
     end
 
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -94,9 +94,10 @@ en:
               inclusion: You must be 16 or over to use this service
               in_the_future: Their date of birth must be in the past
               invalid: Enter their date of birth in the correct format
-              missing_day: Enter a day for their date of birth, formatted as a number
-              missing_month: Enter a month for their date of birth, formatted as a number
-              missing_year: Enter a year with 4 digits
+              missing_day: Date of birth must include a day
+              missing_month: Date of birth must include a month
+              missing_year: Date of birth must include a year
+              incomplete_year: Year must include 4 numbers
             age_known:
               inclusion: Select yes if you know their date of birth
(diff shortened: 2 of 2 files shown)


config/locales/en.yml
```diff
@@ -94,9 +94,9 @@ en:
               inclusion: You must be 16 or over to use this service
               in_the_future: Their date of birth must be in the past
               invalid: Enter their date of birth in the correct format
-              missing_day: Enter a day for their date of birth, formatted as a number
-              missing_month: Enter a month for their date of birth, formatted as a number
-              missing_year: Enter a year with 4 digits
+              missing_day: Date of birth must include a day
+              missing_month: Date of birth must include a month
+              missing_year: Date of birth year must include 4 numbers
```
vassyz, line 99: Is that the correct wording? What about birthyear or year of birth?

paulodeon: Good question, the ticket actually asks for "Year must include 4 numbers" but I took the liberty of deviating from that for consistency. See also the other year fields which need validation

Dan13L3: @paulodeon @vassyz @adamsilver 
I think we need different error messages depending on the error, so:

- [if year is missing completely]: ‘Date of birth must include a year’

- [if year has been entered, but is incomplete]: ‘Year must include 4 numbers’

And then follow that pattern for the other messages.

paulodeon: Ok @Dan13L3 I've changed that now - I had to add a new validation case to the date validator - wdy think?

Dan13L3: Thanks @paulodeon - looks good to me

paulodeon: Great @Dan13L3 could I get an approval please

app/validators/date_validator.rb
```diff
@@ -61,11 +61,16 @@ def validate_each(record, attribute, _value)
       return false
     end
 
-    if year < 1000
+    if year.zero?
       record.errors.add(attribute, :missing_year)
       return false
     end
 
+    if year < 1000
```
tavy88, line 69: Do we want to allow users to input year 1001+ ?

tavy88 (approved):
LGTM

PR #422, opened 2023-01-26
Add referral form docs

paulodeon:
### Context

To make it easier for developers to get started 

Add documentation highlighting the form structure 

### Guidance to review

Does it make sense?
Anything else we should add?

### Link to Trello card

No trello card

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/docs/referrer-form-structure.md
+++ b/docs/referrer-form-structure.md
@@ -1,0 +1,58 @@
+# Referer form structure
+
+The form is made up of:
+
+- **Eligibility screener** - Filters out ineligible referrals
+- **Main Referral form** - capture the referral and send to the team
+
+## Elegibility screener
+
+This takes you through the following steps:
+
+```
+ReferralTypeController#new
+IsTeacherController#new
+TeachingInEnglandController#new
+SeriousMisconductController#new
+```
+
+Each form shown on `new` posts to `create` on the same controller then redirects to the next step.
+
+Once done you're redirected to the final page for this section:
+
+```
+PagesController#you_should_know
+```
+
+# Main referral form
+
+If you're eligible you'll be sent to the referral controller where a new referral object is saved. All further actions are performed on the referral object. You're then sent to the top level form for the referral at `#edit`:
+
+```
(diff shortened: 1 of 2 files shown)


docs/referrer-form-structure.md
```diff
@@ -0,0 +1,62 @@
+# Referer form structure
+
+Once you visit `/start` enter your email address and the OTP sent to your email you'll be sent to the eligibility screener.
```
vassyz, line 3: I think this needs better wording.

docs/referrer-form-structure.md
```diff
@@ -0,0 +1,62 @@
...
+
+This will filter out those applications who are ineligible for this form.
```
vassyz, line 5: Maybe try to say that the applicants only have access to their own forms or something related to authentication and authorisation. Filtering out ineligible applications sounds like something else to me but maybe we're taking about two different things.

docs/referrer-form-structure.md
```diff
@@ -0,0 +1,62 @@
...
+
+Until we show them all the steps and confirm here
+
+```
+Referrals::ReferrersController#show
+```
+
+### Form objects
+
+Each of the steps above has it's own form object so for instance [app/controllers/referrals/referrer_phone_controller.rb](../app/controllers/referrals/referrer_phone_controller.rb) uses [app/forms/referrer_phone_form.rb](../app/forms/referrer_phone_form.rb)
```
vassyz, line 58: Minor typo `its own`.

docs/referrer-form-structure.md
gpeng: There's a ticket to change this flow too that we'll do this sprint so this is going to change in any case. We could probably leave this whole bit out and just have the eligibility check start with the start button. The email signup is going to be moved to the end of the process as per the prototype.

paulodeon: I've addressed that now

vassyz (approved):

PR #423, opened 2023-01-26
Don't delete all of the bundler things

gpeng:
### Context

Gems aren't available when we SSH onto a running instance.

* Don't be quite so aggressive with the tidying up in the image

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


gpeng:
Closing as we have an alternative solution.

PR #427, opened 2023-01-27
Fix zip file download when using an external storage service

vassyz:
### Context

Adding the files to the zip file was working correctly locally (or on a server not using an external service) because the file can be read directly from the filesystem. When using a storage service (currently Azure) the code was throwing an error:

```
NoMethodError undefined method `path_for' for #<ActiveStorage::Service::AzureStorageService:
```

Instead of adding files to the zip file by their paths, it downloads the attachments from the external service to a temp file and adds the data directly to the zip file.
Rails used to have a class that was able to do this (`ActiveStorage::Downloading`), but they [deprecated it](https://github.com/rails/rails/commit/ee21b7c2eb64def8f00887a9fafbd77b85f464f1). I choose to download the blob to disk instead of streaming it into memory because even if it might be slower, I'd rather use the filesystem storage than RAM. I tested it on the development server and with a 10Mb attachment it gets generated in ~4s.

### Link to Trello card

https://trello.com/c/9LhoQTjO/1162-bug-zip-file-file-export-on-azure

### Solves

https://sentry.io/organizations/dfe-teacher-services/issues/3896636999/?project=4504136292237312&referrer=slack

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/lib/referral_zip_file.rb
+++ b/app/lib/referral_zip_file.rb
@@ -22,21 +22,35 @@ class ReferralZipFile
   private
 
   def zip_file
-    temp_file = Tempfile.new("temp.zip")
+    temp_zip_file = Tempfile.new("temp.zip")
+    temp_attachments = []
 
-    Zip::File.open(temp_file.path, create: true) do |zip_file|
+    Zip::File.open(temp_zip_file.path, create: true) do |zip|
       attachments.each do |attachment|
         next unless attachment.attached?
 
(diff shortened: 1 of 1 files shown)


vassyz:
[[BUG] Zip file file export on Azure](https://trello.com/c/9LhoQTjO/1162-bug-zip-file-file-export-on-azure)

gpeng (approved):

PR #439, opened 2023-01-31
Update employer personal details form

felixclack:
For any question that has the option to say no, we want to allow a user
to change their answer.

This means adding a relevant row to the check answers page.

See [the prototype](https://refer-misconduct-prototype.herokuapp.com/report/teacher/check-answers) for reference.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

Screenshot 2023-01-31 at 12 00 21 pm

### DIFF
--- a/app/components/personal_details_component.rb
+++ b/app/components/personal_details_component.rb
@@ -12,7 +12,7 @@ class PersonalDetailsComponent < ViewComponent::Base
     return rows if referral.from_member_of_public?
 
     rows.push(date_of_birth_known_row)
-    rows.push(date_of_birth_row)
+    rows.push(date_of_birth_row) if referral.age_known?
     rows.push(trn_known_row)
     rows.push(trn_row) if referral.trn_known?
     rows.push(qts_row)
@@ -125,7 +125,7 @@ class PersonalDetailsComponent < ViewComponent::Base
         text: "Do you know their date of birth"
       },
       value: {
-        text: referral.date_of_birth&.present? ? "Yes" : "No"
+        text: referral.age_known? ? "Yes" : "No"
       }
     }
   end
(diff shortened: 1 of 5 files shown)


app/components/personal_details_component.rb
```diff
@@ -105,4 +61,115 @@ def name_row
...
+              return_to: request.path
+            ),
+          visually_hidden_text: "if you know their date of birth"
+        }
+      ],
+      key: {
+        text: "Do you know their date of birth"
+      },
+      value: {
+        text: referral.date_of_birth&.present? ? "Yes" : "No"
```
vassyz, line 104: This can cause a bug because I don't think we're making the DOB nil when changing to `No`. That means once you answer `Yes` once it will forever be `Yes` 😄

app/components/personal_details_component.rb
```diff
@@ -5,83 +5,39 @@ class PersonalDetailsComponent < ViewComponent::Base
...
-          }
-        ],
-        key: {
-          text: "Do they have QTS?"
-        },
-        value: {
-          text: referral.has_qts&.humanize
-        }
+      value: {
+        text: referral.name_has_changed&.humanize
```
vassyz, line 38: Should it also show the `Other name` row?.

app/components/personal_details_component.rb
```diff
@@ -5,83 +5,39 @@ class PersonalDetailsComponent < ViewComponent::Base
   attr_accessor :referral
 
   def rows
-    return [name_row] if referral.from_member_of_public?
+    rows = [name_row, any_other_name_row]
```
vassyz, line 8: Not in this PR, but the `name_row` has the title `Their name` in the prototype.

vassyz (approved):
Looks good, just the linter fixes

PR #452, opened 2023-02-06
1180 - Update about you section user journey

tavy88:
### Context

Additional tweaks to the user journey for About you section.

- If the user starts filling the About you section - Your details, but does not get to the check your answers page (e.g does select whether the form is complete or not), one starts from the first question.
- If the user starts filling the About you section - Your organisation, but does not get to the check your answers page (e.g does select whether the form is complete or not), one starts from the first question.

Otherwise, one lands on the check you answers page in both cases

https://user-images.githubusercontent.com/1955084/217013441-c973d406-8006-4424-9e9c-e96aed8533cc.mov

### Link to Trello card

https://trello.com/c/i3wl0AM8

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/forms/referral_form.rb
+++ b/app/forms/referral_form.rb
@@ -56,11 +56,15 @@ class ReferralForm
   end
 
   def about_you_section_path
-    if referral.referrer.present?
-      return polymorphic_path([referral.routing_scope, referral, :referrer])
+    if referral.referrer&.complete.nil?
+      return(
+        polymorphic_path(
+          [:edit, referral.routing_scope, referral, :referrer_name]
+        )
+      )
     end
 
-    polymorphic_path([:edit, referral.routing_scope, referral, :referrer_name])
+    polymorphic_path([referral.routing_scope, referral, :referrer])
   end
 
   def about_the_person_you_are_referring_section
--- a/app/forms/referral_form.rb
+++ b/app/forms/referral_form.rb
@@ -267,12 +267,14 @@ class ReferralForm
   end
 
   def about_your_organisation_section_path
-    if referral.organisation.present?
-      return polymorphic_path([referral.routing_scope, referral, :organisation])
+    if referral.organisation&.complete.nil?
+      return(
+        polymorphic_path(
+          [:edit, referral.routing_scope, referral, :organisation_address]
+        )
+      )
     end
 
-    polymorphic_path(
-      [:edit, referral.routing_scope, referral, :organisation_address]
-    )
+    polymorphic_path([referral.routing_scope, referral, :organisation])
   end
 end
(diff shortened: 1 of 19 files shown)


app/forms/referral_form.rb
```diff
@@ -263,12 +267,14 @@ def path_for_section_status(status, start_path, check_answers_path)
   end
 
   def about_your_organisation_section_path
-    if referral.organisation.present?
-      return polymorphic_path([referral.routing_scope, referral, :organisation])
+    if referral.organisation&.complete.nil?
```
vassyz, line 270: Not sure why the order was reversed here, I don't think it makes a difference when replacing the previous conditional with `referral.organisation&.complete.present?`

spec/forms/organisation_form_spec.rb
```diff
@@ -34,7 +34,7 @@
 
     context "when the value of complete is nil" do
       let(:complete) { nil }
-      let(:organisation) { build(:organisation, completed_at: Time.current) }
+      let(:organisation) { build(:organisation) }
```
vassyz, line 37: This whole `describe "#complete" do` can be removed I think.

spec/forms/referrer_form_spec.rb
```diff
@@ -21,7 +21,7 @@
     let(:referrer_form) { described_class.new(complete:, referrer:) }
 
     context "when the value of complete is nil" do
-      let(:referrer) { build(:referrer, completed_at: Time.current) }
+      let(:referrer) { build(:referrer) }
```
vassyz, line 24: Same here, this method doesn't exist anymore.

vassyz (approved):
Looks good, just a few minor observations.

app/forms/referral_form.rb
tavy88:
```ruby
referral.organisation&.complete.present?
```
wouldn't work because if `referral.organisation&.complete == false` this wouldn't go to 
```ruby
 polymorphic_path(
          [:edit, referral.routing_scope, referral, :organisation_address]
        )
```
But if want it to. We just don't want it to happen when complete is `nil`.

spec/forms/organisation_form_spec.rb
tavy88: I guess it's testing the `attr_reader :complete`. Not much value but thought I'd keep it. If you feel strongly about it I will remove.

app/forms/referral_form.rb
vassyz: You're right. Sorry, missed that.

vassyz (approved):

vassyz:
[[BUG] Referrer completed `false` value can't be saved](https://trello.com/c/tVuMeL3r/1106-bug-referrer-completed-false-value-cant-be-saved)

PR #454, opened 2023-02-06
[DO NOT MERGE] Spike PDF creation

gpeng:
### Context

We need to generate a PDF version of a referral that can be redacted and distributed. The design bundles this document in a ZIP file with the other file attachments for the case workers to download. The ZIP file functionality is already done but it doesn't have the PDF as yet.

This PR is a couple of approaches to generating the PDF. Both approaches use Grover to generate a PDF from an HTML page as this seemed the least effort. There is also wkhtmltopdf but that seems pretty much the same sort of thing. 

### Approach 1

Generate the PDF in a controller action from the standard show page. (in the first commit). The implementation here would require another request to be made within the request for the ZIP file. I did try just rendering the templates directly but there are lot of things the layout requires that we'd need to stub.

### Approach 2

Generate the PDF in a job. I've used a separate template for this that will need a bit of design input but it saves a lot of the stubbing that would be needed to use the existing layouts. It does use the existing components but I couldn't get the path helpers working so they would need to be refactored a bit like I've done with the organisation component here.

I think we should go with something like approach 2 and trigger the job when the referral is submitted. We can attach the file to the referral and then the PDF will be included in the ZIP file. I haven't included that bit in the spike.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


gpeng:
[SPIKE: PDF generation options](https://trello.com/c/tvI8nA17/1123-spike-pdf-generation-options)

gpeng:
Closing now as this has been built in #478

PR #459, opened 2023-02-07
Remove validate_inclusion_of for booleans

vassyz:
The use of `validate_inclusion_of` for booleans is [discouraged](https://www.rubydoc.info/gems/shoulda-matchers/5.1.0/Shoulda/Matchers/ActiveModel).

```
We discourage using `validate_inclusion_of` with boolean columns. In fact, there is never a case where a boolean column will be anything but true, false, or nil, as ActiveRecord will type-cast an incoming value to one of these three values. That means there isn't any way we can refute this logic in a test. Hence, this will produce a warning.
```

### DIFF
--- a/spec/forms/referrals/teacher_role/end_date_form_spec.rb
+++ b/spec/forms/referrals/teacher_role/end_date_form_spec.rb
@@ -9,12 +9,6 @@ RSpec.describe Referrals::TeacherRole::EndDateForm, type: :model do
 
   describe "validations" do
     it { is_expected.to validate_presence_of(:referral) }
-
-    specify do
-      expect(form).to validate_inclusion_of(:role_end_date_known).in_array(
-        [true, false]
-      )
-    end
   end
 
   describe "#valid?" do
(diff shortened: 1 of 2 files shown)


spec/forms/referrals/teacher_role/end_date_form_spec.rb
```diff
@@ -9,12 +9,6 @@
 
   describe "validations" do
     it { is_expected.to validate_presence_of(:referral) }
-
-    specify do
-      expect(form).to validate_inclusion_of(:role_end_date_known).in_array(
-        [true, false]
-      )
-    end
   end
```
tavy88, line 12: Are you thinking of dropping the validations at the model level as well? There's plenty of models that use this from what I've seen. Also are these specs the only ones?

vassyz: It's only the ones that test booleans.

tavy88: Yeah I think it's this one too: https://github.com/DFE-Digital/refer-serious-misconduct/blob/main/spec/forms/referrals/teacher_role/end_date_form_spec.rb#L15

tavy88 (approved):

PR #460, opened 2023-02-07
1081 - User sees only COMPLETE/INCOMPLETE statuses

tavy88:
### Context

 User sees only COMPLETE/INCOMPLETE statuses

On the task-list user only sees COMPLETE/INCOMPLETE statuses for a better UX. However, we do keep the logic behind that relies on three statuses currently:
- :complete
- :incomplete
- :not_started_yet

These are still needed to ensure correct service behavior.

### Guidance to review

Screenshot 2023-02-07 at 14 51 22

### Link to Trello card

https://trello.com/c/VadMIhzi

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/components/task_list_section_component.html.erb
+++ b/app/components/task_list_section_component.html.erb
@@ -11,7 +11,7 @@
         </span>
         <% tag_css = item.status != :completed ? " govuk-tag--grey" : "" -%>
         <strong class="govuk-tag app-task-list__tag<%= tag_css %>" id="<%= item.label.parameterize %>">
-          <%= item.status.to_s.humanize %>
+          <%= I18n.t("referral_form.statuses.#{item.status}") %>
         </strong>
       </li>
     <% end %>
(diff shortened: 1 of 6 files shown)


vassyz:
If it desn't serve any purpose, shouldn't we remove the `not_started_yet` status?

tavy88:
> If it desn't serve any purpose, shouldn't we remove the `not_started_yet` status?

We can't really do that without massive work because our entire user journey logic and where to land from the task list relies on the `not_started_yet` status. We still need that logic, but for the user it still means incomplete. We can discuss a potential refactor if we want to get rid completely of `not_started_yet`.

vassyz (approved):
I'll approve it, because as you say, removal of the `not_started_yet` requires some work. I might have a go at refactoring as the status doesn't have a consistent logical meaning anymore.

PR #461, opened 2023-02-08
Refactor section status

vassyz:
The section status was responsible for two different things from a
logical perspective: setting the section link to go to the check
answers page and display the status badge for the section row. With
the removal of the `not_started_yet` status these logical checks weren't
aligning anymore.


tavy88 (approved):
Looks great! Let's check the deployed version to make sure it works as expected.

PR #467, opened 2023-02-09
Remove sign-in/out related flash messages

felixclack:
We no longer want to display the flash messages around authentication.

We initially wrote specific logic to ensure they are displayed. Rather
than attempt to overwrite Devise to prevent them being generated, I
opted to simply clear the values of the messages.

This gives us the flexibility to use the other parts of the Devise flash
messages while removing the specific sign-in/out ones.

### Link to Trello card

https://trello.com/c/1W0sWBOh/1186-remove-sign-in-sign-out-flash-messages

https://www.loom.com/share/886fe4ddcb0e4e178fea5d3398c22b03

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/components/personal_details_component.rb
+++ b/app/components/personal_details_component.rb
@@ -76,7 +76,7 @@ class PersonalDetailsComponent < ViewComponent::Base
               [:edit, referral.routing_scope, referral, :personal_details_name],
               return_to: request.path
             ),
-          visually_hidden_text: "their other name"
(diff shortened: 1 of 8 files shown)


vassyz:
I think this will also hide the login error messages, like invalid password.

vassyz (approved):

PR #472, opened 2023-02-09
1184 restrict access case workers

tavy88:
### Context

Add case worker roles

We are introducing two roles for the staff users:
- view_support
- manage_referrals

Staff users with view_support role will have access to /support/*
Staff users with manage_referrals will have access to manage/*

Users authenticated through basic auth will no longer be
authorized to access the mentioned routes.

Roles management from the UI is out of this PR's scope.

Screenshot 2023-02-09 at 13 50 53

### Note

This is a breaking change as it restricts access to /support and /manage to the existing case workers, unless they have the corresponding flags turned on.

### Link to Trello card

https://trello.com/c/5XGFG4MQ

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


app/controllers/support_interface/support_interface_controller.rb
```diff
@@ -1,8 +1,21 @@
...
+
     layout "support_layout"
 
     before_action :authenticate_staff!
+    after_action :verify_authorized
+
+    private
+
+    def staff_user_not_authorized
+      flash[:warning] = "You are not authorized to access this section."
```
tavy88, line 16: This message is flexible of course.

spec/system/manage/case_worker_views_a_public_referral_spec.rb
```diff
@@ -4,15 +4,26 @@
 RSpec.feature "Manage referrals" do
   include CommonSteps
 
-  scenario "Case worker views a referral" do
+  scenario "Case worker with basic auth is not authorized to see referrals" do
```
malcolmbaig, line 7: I think we can safely skip testing the staff basic auth scenarios in our system specs. Reasoning:

- This functionality only exists as a way of getting Staff users set up in an environment. Now that production has these users set up, the basic auth isn't something the live app should surface again
- We should move towards using actual Staff user authentication in our specs, rather than using the basic auth as a crutch

As an aside, we should probably consider making staff basic auth either dev-only, or add something to https://github.com/DFE-Digital/refer-serious-misconduct/blob/main/app/lib/staff_http_basic_auth_strategy.rb#L3 to ensure it can't be switched on in prod. @gpeng WDYT?

spec/system/manage/case_worker_views_a_public_referral_spec.rb
```diff
@@ -4,15 +4,26 @@
...
     and_the_eligibility_screener_feature_is_active
     and_staff_http_basic_is_active
-    and_i_am_authorized_as_a_case_worker
+    and_i_am_authorized_with_basic_auth_as_a_case_worker
     and_there_is_an_existing_public_referral
 
     when_i_visit_the_referral
+    then_i_am_unauthorized_and_redirected_to_root_path
+  end
+
```
malcolmbaig, line 18: A general observation that applies to all of these system spec changes:

Our [testing styleguide](https://github.com/DFE-Digital/refer-serious-misconduct/blob/main/docs/testing_styleguide.md) specifies that we have one scenario per file. One option is to split these out, (eg—the scenario below would live in a file called `case_worker_with_permissions_views_a_public_referral_spec.rb`).

Alternatively, it might be sufficient to cover this behaviour in a single scenario. Along the lines of:
```
when_i_am_authorized_as_a_case_worker_without_management_permissions
and_i_visit_the_referral
then_i_am_unauthorized_and_redirected_to_root_path

when_i_am_given_management_permissions # a database change to toggle the appropriate field
and_i_visit_the_referral
then_i_see_the_referral_summary
# etc...
```

Similarly for the other system spec files.

spec/support/system/common_steps.rb
```diff
@@ -98,6 +98,67 @@ def and_i_see_personal_details_flagged_as_incomplete
     end
   end
 
+  def when_i_am_authorized_as_a_case_worker_with_management_permissions
+    Capybara.reset_sessions!
```
malcolmbaig, line 102: Doesn't this happen between scenarios anyway? In which case, is it necessary here?

spec/support/system/common_steps.rb
```diff
@@ -98,6 +98,67 @@ def and_i_see_personal_details_flagged_as_incomplete
...
+
+    create(:staff, :confirmed, :can_view_support, :can_manage_referrals)
+
+    visit new_staff_session_path
+
+    fill_in "staff-email-field", with: "test@example.org"
+    fill_in "staff-password-field", with: "Example123!"
+
+    click_button "Log in"
```
malcolmbaig, line 111: Are we able to just `sign_in(@staff)`?

If we wanted to exercise the login screen, we could write a single scenario just for that and use the Devise helper elsewhere.

app/views/support_interface/staff/index.html.erb
```diff
@@ -38,5 +38,14 @@
...
       row.value { staff.last_sign_in_at.to_s }
     end
+
+    roles = []
+    roles << 'view_support' if staff.view_support?
+    roles << 'manage_referrals' if staff.manage_referrals?
+
+    summary_list.row do |row|
+      row.key { 'Roles' }
+      row.value { roles.join(',') if roles.size > 0 }
```
vassyz, line 48: You can add a space after the comma `join(', ')`. Why do you need to check `if roles.size > 0`?
Not in this PR, but we can format the dates on this page with `.to_fs(:day_month_year_time)`.

malcolmbaig:
Hi @tavy88 👋🏼 . Implementation looks good, have left some comments around test structure based on the contents of our testing styleguide.

malcolmbaig:
Do we have a plan for giving the right permissions to existing users on test, preprod and prod?

spec/system/manage/case_worker_views_a_public_referral_spec.rb
tavy88: Right ok.

spec/support/system/common_steps.rb
tavy88: I was just ensuring the session is clean. If a previous step logs a user and then this is the next step, then the session will not be empty. I can try to remove it and see if the tests still pass.

spec/support/system/common_steps.rb
tavy88: Good point.

app/views/support_interface/staff/index.html.erb
tavy88: Yeah you're right it might be an overkill condition.

tavy88:
> Do we have a plan for giving the right permissions to existing users on test, preprod and prod?

Yep. Once it is deployed we will access the CLI and do that manually. I don't currently have access to @vassyz does so he can assist me when the time comes.

spec/system/manage/case_worker_views_a_public_referral_spec.rb
tavy88: Yeah I would keep the tests to ensure the access is not granted by accident to basic auth users for the scope of this PR. That's why I kept it intentionally.

malcolmbaig:
> Yep. Once it is deployed we will access the CLI and do that manually. I don't currently have access to @vassyz does so he can assist me when the time comes.

Suggestion: we could also add a data migration which retrieves Staff records using a list of hard-coded email addresses. Saves having to manually do these changes across three environments.

spec/system/manage/case_worker_views_a_public_referral_spec.rb
gpeng: Yep. I agree 👍 I'll do a ticket.

tavy88:
> > Yep. Once it is deployed we will access the CLI and do that manually. I don't currently have access to @vassyz does so he can assist me when the time comes.
> 
> Suggestion: we could also add a data migration which retrieves Staff records using a list of hard-coded email addresses. Saves having to manually do these changes across three environments.

I think this PR is big enough as it is. I think a simple update statement for the existing users will work for now. We have in plan to create a management UI at a later date.

spec/support/system/common_steps.rb
```diff
@@ -98,6 +98,39 @@ def and_i_see_personal_details_flagged_as_incomplete
...
+
+  def when_i_am_authorized_with_basic_auth_as_a_case_worker
+    page.driver.basic_authorize(
+      ENV.fetch("SUPPORT_USERNAME", "test"),
+      ENV.fetch("SUPPORT_PASSWORD", "test")
+    )
+  end
+  alias_method :and_i_am_authorized_with_basic_auth_as_a_case_worker,
+               :when_i_am_authorized_with_basic_auth_as_a_case_worker
+
```
malcolmbaig, line 133: There's potentially a case for extracting these authorization methods into a module, but I don't think CommonSteps is the appropriate home for them. We should try and limit the steps in this module to things which are very generic and applicable across nearly any section of the app. For now, it would be totally fine to duplicate these wherever they're used more than once.

spec/system/manage/case_worker_with_basic_auth_is_not_authorized_to_view_public_referrals_spec.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+RSpec.feature "Manage referrals" do
+  include CommonSteps
+
+  before do
+    create_list(:referral, 30, :submitted)
+    create_list(:referral, 20)
+  end
```
malcolmbaig, line 10: This should go in a `given_` step, our styleguide discourages the use of `before` blocks in system specs.

spec/support/system/common_steps.rb
tavy88: Agreed.

spec/system/manage/case_worker_with_basic_auth_is_not_authorized_to_view_public_referrals_spec.rb
tavy88: Yeah it's a legacy thing. I'm not even sure we need to create that many referral applications.

app/controllers/support_interface/support_interface_controller.rb
```diff
@@ -1,8 +1,29 @@
...
+
+    private
+
+    def staff_user_not_authorized
+      flash[:warning] = I18n.t("pundit.unauthorized")
+
+      # FeatureFlags::Engine is a mounted engine and we can't control the
+      # behaviour from this level. Attempting to do the redirect
+      # here ends up in multiple redirects.
+      redirect_to root_path unless is_a?(FeatureFlags::FeatureFlagsController)
```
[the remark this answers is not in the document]
tavy88: It happens when I do a manual test actually.

PR #477, opened 2023-02-10
Email to the case worker inbox on submission

vassyz:
### Context

Case worker should get notified when a new referral gets submitted.

Test email preview:

Screenshot 2023-02-13 at 15 58 56

- add a new mailer 
- remove unused header values from application mailer
- set the mailer constants in the application mailer
- add the mailer view
- add a required env `GOVUK_NOTIFY_MANAGE_SERIOUS_MISCONDUCT_API_KEY`
- add an optional env `GOVUK_NOTIFY_MISCONDUCT_TEACHER_EMAIL`

### Link to Trello card

https://trello.com/c/0OrT7FQ1/1113-email-to-the-case-worker-inbox-on-submission

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


vassyz:
[Email to the case worker inbox on submission](https://trello.com/c/0OrT7FQ1/1113-email-to-the-case-worker-inbox-on-submission)

app/mailers/caseworker_mailer.rb
```diff
@@ -0,0 +1,18 @@
...
+  include Rails.application.routes.url_helpers
+
+  before_action :set_action_mailer_misconduct_key
+
+  def referral_submitted(referral)
+    @name = referral.referrer_name
+    @link = manage_interface_referral_url(referral)
+
+    mailer_options = {
+      to: MISCONDUCT_NOTIFY_MAILER_TO,
```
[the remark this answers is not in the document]
vassyz: For now, the requirement is to send all submitted referrals to that specific email address, I added it as an optional env variable so I don't spam that inbox if I have to test it.

app/mailers/application_mailer.rb
```diff
@@ -2,9 +2,19 @@
...
+  MISCONDUCT_NOTIFY_MAILER_TO =
+    ENV.fetch(
+      "GOVUK_NOTIFY_MISCONDUCT_TEACHER_EMAIL",
+      "misconduct.teacher@education.gov.uk"
+    )
+
+  private
+
+  def set_action_mailer_misconduct_key
+    ActionMailer::Base.notify_settings[:api_key] = ENV.fetch(
```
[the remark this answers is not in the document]
tavy88: I would have thought that we have different templates for different envs.

app/mailers/application_mailer.rb
```diff
@@ -2,9 +2,19 @@
 class ApplicationMailer < Mail::Notify::Mailer
   GENERIC_NOTIFY_TEMPLATE = "04b96c89-77ec-4c07-af01-3c96a4b8b5d7"
 
-  def notify_email(headers)
-    headers =
-      headers.merge(rails_mailer: mailer_name, rails_mail_template: action_name)
-    view_mail(GENERIC_NOTIFY_TEMPLATE, headers)
+  MISCONDUCT_GENERIC_NOTIFY_TEMPLATE = "4f8e7de9-3987-4e75-974f-ac94068c4a62"
+  MISCONDUCT_NOTIFY_REPLY_TO_ID = "83cac27e-b914-46a3-a1f9-56e5acb07d05"
```
tavy88, line 6: I would have thought that we have different templates for different envs. Also what if for some reason the current templates in Notify get destroyed? I feel like all these should be set as env vars.

vassyz: We don't have different templates. How would having it as an env var help when the template gets deleted?

tavy88: Because we wouldn't need to do any code changes and deploy. We would just need to update those env vars.

gpeng: I think generally on the other services we've had one template and added the content in the service so it wouldn't change and we don't really need the env var.

tavy88 (approved):

PR #478, opened 2023-02-10
[1110] case worker pdf referral generation

gpeng:
### Context

Case workers require a PDF version of a referral that they can redact and disseminate to panelists. We want to provide this as part of their ZIP bundle that can be downloaded from the referral page.

* Add a RenderPdfJob that can attach a PDF version to the referral
* Add Referral#pdf to attach it to
* Queue up the job to do the PDF on submission of the referral

### Guidance to review

A PDF can be generated and attached to a referral locally by calling `RenderPdfJob.perform_later(referral: <a referral>) in the console. Viewing that referral in the case worker UI and downloading the ZIP bundle. The PDF should be in there.

I've reviewed the PDF design with @adamsilver . There is possibly some work to do to iterate on that it's ok for now.

This is working locally but I think we might need to install some things e.g. chrome headless, on the worker boxes. I'll address that when I've checked on the review app.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,7 +6,6 @@
 .yarn-integrity
 config/master.key
 log
-node_modules
 public/assets
 public/packs
 public/packs-test
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,5 @@ tmp/pids/*
 !tmp/.keep
 !tmp/pids/
 !tmp/pids/.keep
+
+.cache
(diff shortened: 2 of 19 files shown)


gpeng:
[Case worker PDF referral generation](https://trello.com/c/tv2m6mBC/1110-case-worker-pdf-referral-generation)

tavy88 (approved):
LGTM

tavy88 (approved):

PR #482, opened 2023-02-13
Fix README console docs

gpeng:
Minor change to the review app console instructions.


malcolmbaig (approved):

vassyz (approved):
I was about to fix that in my current PR. I had a separate command initially for review apps, but ended up bundling all envs into one.

PR #490, opened 2023-02-15
1038 -  Big query integration

tavy88:
### Context

Add event tracking using BigQuery

The types of events we are tracking are:

- web_request - sent after a controller action is performed using controller callbacks
- create_entity - sent after an object is created using model callbacks
- update_entity - sent after an object is updated using model callbacks
- delete_entity - sent after an object is deleted using model callbacks
- import_entity - sent for each object imported using the DfE Analytics import rake tasks

Also, define which data fields are:

- allowed to be tracked
- blocked from tracking
- classified as pii

Add event tracking using [BigQuery](https://github.com/DFE-Digital/dfe-analytics)

Screenshot 2023-03-07 at 11 29 48

Screenshot 2023-03-07 at 11 29 59

### Guidance to review

Please read the documentation about how to get started with [dfe-analytics](https://github.com/DFE-Digital/dfe-analytics)

### Link to Trello card

https://trello.com/c/Ts0YO2nH

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


gpeng (changes requested):
I think we're missing quite a few fields from the PII. The long text details fields will almost certainly contain PII and I think we'd probably want to include work and organisation address fields from referral. Also the fields from `organisations`. Not sure why we have that at all but while we have we should include those.

gpeng (approved):

PR #491, opened 2023-02-15
Rate limit OTP emails

malcolmbaig:
### Context

The recent pentest identified a need to rate limit requests to the endpoint that generates and emails out the OTP for user sign in.

- Add Rack::Attack
- Add a throttle for POST requests to `/users/session` 

### Guidance to review
The spec outlines the behaviour, but this could also be manually tested by hitting the rate limit with curl.

### Link to Trello card
https://trello.com/c/Efy19xWe/1188-rate-limit-confirmation-code-emails

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


config/initializers/rack_attack.rb
```diff
@@ -0,0 +1,3 @@
+Rack::Attack.throttle("limit_otp_emails", limit: 10, period: 1.minute) do |req|
```
malcolmbaig, line 1: For posterity — I opted for 10 per minute as that seems like a reasonable number at which requests to this endpoint from a single IP start to look unusual. We may need to keep an eye on things and adjust later.

PR #495, opened 2023-02-16
Add fallback redis URL in production cache config

malcolmbaig:
Needed in the deploy pipeline when building the app, the actual deployed environments have a REDIS_URL configured.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -62,7 +62,9 @@ Rails.application.configure do
   config.log_tags = [:request_id]
 
   # Use a different cache store in production.
-  config.cache_store = :redis_cache_store, { url: ENV.fetch("REDIS_URL") }
+  config.cache_store =
+    :redis_cache_store,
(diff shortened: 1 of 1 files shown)


config/environments/production.rb
```diff
@@ -62,7 +62,7 @@
   config.log_tags = [:request_id]
 
   # Use a different cache store in production.
-  config.cache_store = :redis_cache_store, { url: ENV.fetch("REDIS_URL") }
+  config.cache_store = :redis_cache_store, { url: ENV.fetch("REDIS_URL", "redis://localhost:6379/1") }
```
tavy88, line 65: Is this correct? 

`redis://localhost:6379/1` 

Is it not: `redis://localhost:6379` ?

tavy88 (approved):

vassyz (approved):

config/environments/production.rb
malcolmbaig: That would default to the `/0` database on this instance. We don't really have a need to use multiple DBs, so just using the same one that we've used elsewhere.

PR #497, opened 2023-02-16
Refactor support specs

tavy88:
### Context

Move some of the common steps to their own module.

Move some of the common steps to their own module.


spec/support/system/activate_feature_steps.rb
```diff
@@ -0,0 +1,21 @@
+module ActivateFeatureSteps
```
vassyz, line 1: Should this be plural? feature*s*

vassyz (approved):
Looks good 👍

spec/support/system/activate_feature_steps.rb
tavy88: Sure

PR #499, opened 2023-02-17
1213 - User sees new copy when the security code has expired

tavy88:
### Context

User sees new copy when the security code has expired

Screenshot 2023-02-17 at 11 19 26

### Link to Trello card

https://trello.com/c/g3XRtxGo

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/views/users/otp/retry.html.erb
+++ b/app/views/users/otp/retry.html.erb
@@ -1,18 +1,18 @@
-<h1 class="govuk-heading-l">There was a problem signing in</h1>
+<h1 class="govuk-heading-l"><%= t("users.retry.#{@error}.heading") %></h1>
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <% case @error %>
     <% when 'expired' %>
       <p>
-        <%= t('users.retry.expired') %>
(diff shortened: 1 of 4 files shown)


app/views/users/otp/retry.html.erb
```diff
@@ -1,18 +1,19 @@
...
-      </p>
-    <% end %>
+    <p>
+      <% if @error == 'expired' %>
+        <p>
+          <%= t("users.retry.#{@error}.body_html", href: govuk_link_to('request a new confirmation code', new_user_session_path(new_referral:), method: :get)) %>
+        </p>
+      <% else %>
+        <p>
+          <%= t("users.retry.#{@error}.body") %>
```
vassyz, line 12: I'd personally not use interpolation for `@error` as it can't dynamically have a different value. The else branch makes it look like it can catch what's not `expired`, but it can only be `exhausted`. The `case` statement it replaces looks better IMO, especially if we were to add another error type in the future.

vassyz (commented):
Minor comment, but looks good.

vassyz (approved):

app/views/users/otp/retry.html.erb
tavy88: The problem with the if was that the `expired` page is slightly different (so far) than the rest of potential errors. I hated it to add the `if` tbh. We would still need to interpolate at the headings though.

PR #504, opened 2023-02-20
1207 - Allow videos(mp4, mov) to be uploaded

tavy88:
### Context

 Allow videos(mp4, mov) to be uploaded
 Also adjust the max file size to 50MB.

 Allow users to upload videos(mp4, mov) as evidence. Also increase the max file size to 50MB.

Screenshot 2023-02-20 at 09 02 49

### Link to Trello card

https://trello.com/c/HbMGSRS2

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/validators/file_upload_validator.rb
+++ b/app/validators/file_upload_validator.rb
@@ -1,7 +1,7 @@
 class FileUploadValidator < ActiveModel::EachValidator
   include ActionView::Helpers::NumberHelper
 
-  MAX_FILE_SIZE = 25.megabytes
+  MAX_FILE_SIZE = 50.megabytes
   MAX_FILES = 10
 
   CONTENT_TYPES = {
@@ -16,6 +16,8 @@ class FileUploadValidator < ActiveModel::EachValidator
     ".jpg" => "image/jpeg",
     ".jpeg" => "image/jpeg",
     ".mp3" => "audio/mpeg",
+    ".mp4" => "video/mp4",
+    ".mov" => "video/quicktime",
     ".pdf" => "application/pdf",
     ".png" => "image/png",
     ".txt" => "text/plain",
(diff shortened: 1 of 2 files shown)


app/validators/file_upload_validator.rb
```diff
@@ -16,6 +16,8 @@ class FileUploadValidator < ActiveModel::EachValidator
     ".jpg" => "image/jpeg",
     ".jpeg" => "image/jpeg",
     ".mp3" => "audio/mpeg",
+    ".mp4" => "video/mp4",
+    ".mov" => "video/quicktime",
```
vassyz, line 20: Just these two? A few other formats that come to mind:  .avi , .divx , .mpg.

spec/validators/file_upload_validator_spec.rb
```diff
@@ -35,7 +35,7 @@
   context "with a large file" do
     let(:files) { fixture_file_upload("upload1.pdf", "application/pdf") }
 
-    before { allow(files).to receive(:size).and_return(25 * 1024 * 1024) }
+    before { allow(files).to receive(:size).and_return(51 * 1024 * 1024) }
```
paulodeon, line 38: Maybe consider `50.megabytes + 1` here

app/validators/file_upload_validator.rb
tavy88: Yes if you look at the ticket I think that's what we agreed with @NickGraham101

spec/validators/file_upload_validator_spec.rb
tavy88: What would the advantage be? This test proves that whatever is over the limit makes the object invalid.

paulodeon: I think `50.megabytes + 1` or indeed `51.megabytes` is more readable, but I don't feel strongly about it!

gpeng (approved):

PR #513, opened 2023-02-22
Email user a link to referral when it's created

paulodeon:
### Context

To allow users to access their referral if they partially complete it

https://tra-digital-design-history.herokuapp.com/teacher-misconduct/sending-users-a-link-to-access-their-referral/

* Add referral link email
* Add callback to send email when creating referral
* Rename `send_otp` for consistency with the new mailer function
* Added tests

### Link to Trello card

[](https://trello.com/c/iud9pIjs/1210-referrals-design-history-change-email-user-a-link-to-their-referral-when-started)

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


app/models/user.rb
```diff
@@ -37,11 +37,13 @@ def create_referral!(eligibility_check_id:)
     referrals.create!(eligibility_check_id: eligibility_check.id)
   end
 
-  def find_or_create_referral!(eligibility_check_id:)
+  def find_or_create_referral!(eligibility_check_id:, &block)
     eligibility_check = EligibilityCheck.find_by!(id: eligibility_check_id)
+
     Referral.find_or_create_by!(
```
gpeng, line 43: This doesn't feel like it belongs in here to me. We're making the `User` need to know about referrals and mails and eligibility checks. What about `ReferralsService::Create.call(stuff:...` that does all of the things?

gpeng (changes requested):

app/controllers/referrals_controller.rb
```diff
@@ -6,8 +6,10 @@ class ReferralsController < Referrals::BaseController
   def new
     new_referral =
       current_user.find_or_create_referral!(
-        eligibility_check_id: session.delete(:eligibility_check_id)
+        eligibility_check_id: session.delete(:eligibility_check_id),
+        &:send_referral_link_to_user
```
vassyz, line 10: Why are you sending it as a block? I'd call the mailer with the `new_referral` argument instead of user. Selecting the `latest_referral` instead of the newly created referral seems not ideal to me.

app/models/user.rb
paulodeon: The user has a `has_many :referrals` so referral functionality within the user object seems reasonable to me. This function was already here and all I added was the ability to pass a block (see other comment). Also there was no existing service objects in the project. Given that, this would be my preferred solution in this case, however I'm also happy to add a service object here.

app/controllers/referrals_controller.rb
paulodeon: Rails `find_or_create_by` takes a block which is passed to create if the record is created. I have made the `find_or_create_referral!` method consistent with that. If the record is created the `send_referral_link_to_user` method is called on the newly created record.

Good point on `latest_referral` fixed.

paulodeon: Actually I take your other point as well

app/models/user.rb
paulodeon: I've put the mailer call into the controller which is consistent with the other mailer calls in the app

tavy88: This method seems a bit odd as I don't think `User` should know about `Eligibility`. Feels like `Referral` should know about it. Wonder if you could do something to limit the knowledge of `User` just to `referral` here?

paulodeon: Ok, it seems that wasn't liked! I removed that from the `User` model and it all sits at the controller level now, very clear, consistent with the rest of the app and easy to extract to a service object should we want to later

app/controllers/referrals_controller.rb
```diff
@@ -5,17 +5,29 @@ class ReferralsController < Referrals::BaseController
...
-      current_user.find_or_create_referral!(
+      Referral.find_or_create_by!(
+        user_id: current_user.id,
         eligibility_check_id: session.delete(:eligibility_check_id)
       )
+
+    if new_referral.persisted?
+      UserMailer.referral_link(new_referral).deliver_later
+    end
+
```
tavy88, line 16: Much better!

tavy88 (approved):
LGTM

app/controllers/referrals_controller.rb
```diff
@@ -5,17 +5,29 @@ class ReferralsController < Referrals::BaseController
 
   def new
```
gpeng, line 6: This action seems weird that we're doing all of this in a GET request. I realise it didn't get introduced here, just sayin'.

gpeng (approved):

app/controllers/referrals_controller.rb
paulodeon: I think the create is unused here as well but let's clean this up later.

It is idempotent and it is accessed through a redirect from the otp controller (we need a get for that) so I think it's reasonable

PR #517, opened 2023-02-23
Reorganise referral referrer form objects 

paulodeon:
### Context

The referral/referrer forms, controllers and views are in the main view folder unlike other sections

The same problem exists for the organisation section

Also the naming of the check answers controller/form/view is not consistent for those sections

I'm guessing these were built earlier perhaps by someone else

Finally I discovered a bug which I also fixed, see more here: https://www.loom.com/share/f3befa96dde5480085dd5bbd4ec28be4

Making routing, naming and file location of forms, controllers, views and check answers consistent for the organisation and referrer sections with the rest of the sections.

Create a return_to helper for use in the component views which returns to the correct view, even in the edge case described in the video above

### Checklist

- [ ] Rebased main
- [ ] Cleaned commit history


app/controllers/referrals/referrer_job_title_controller.rb
```diff
@@ -2,12 +2,12 @@ module Referrals
   class ReferrerJobTitleController < Referrals::BaseController
     def edit
       @referrer_job_title_form =
-        ReferrerJobTitleForm.new(referral: current_referral)
+        Referrals::Referrer::JobTitleForm.new(referral: current_referral)
```
tavy88, line 5: Would it not make sense to have 

```ruby
app/controllers/referrers/job_title_controller.rb
```

Just wondering...

paulodeon: I think this makes sense, all the other form sections were in a subfolder so I moved this section into one for consistency. The `Refferer` portion of the form is in fact filling in the `Referrer` object as well so that is consistent

paulodeon:
All of these controllers, routes and forms are already tested, errors were found on all the changes and fixed. 

I'm fairly confident in the test suite in place and would recommend a couple of us do some manual testing as a sanity check.

Otherwise I'm fundamentally not actually changing functionality here, just renaming things...

paulodeon:
I'm going to break this down into more manageable PRs - hopefully that will accelerate getting it merged

paulodeon:
This has now been fully merged with the other PRs

PR #521, opened 2023-02-24
1221 - Improve user flow for case workers

tavy88:
### Context

Improve user flow for case workers as follows:

- if the case worker has only view_support permission, when logging in one gets directed to Staff section
- if case worker has only manage_referrals permission, when logging in one gets directed to manage referrals section
- if case worker has both permissions, one gets directed to manage referrals section, but sees the support sections as well on the navigation

https://user-images.githubusercontent.com/1955084/221222944-57b8a6bf-5e96-4d50-a7f2-a26e8a5f9de0.mov

### Link to Trello card

https://trello.com/c/h4RYonlR/1221-staff-login-and-navigation

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/staff/sessions_controller.rb
+++ b/app/controllers/staff/sessions_controller.rb
@@ -2,6 +2,8 @@
 
 class Staff::SessionsController < Devise::SessionsController
   # before_action :configure_sign_in_params, only: [:create]
+  skip_before_action :require_no_authentication, only: %i[new]
+  before_action :check_signed_in, only: %i[new]
 
   # GET /resource/sign_in
   # def new
@@ -18,10 +20,29 @@ class Staff::SessionsController < Devise::SessionsController
   #   super
   # end
 
+  def after_sign_in_path_for(_resource)
+    if current_staff.manage_referrals?
+      manage_interface_referrals_path
+    elsif current_staff.view_support?
+      support_interface_staff_index_path
+    else
+      flash[:warning] = I18n.t("pundit.unauthorized")
+      root_path
+    end
+  end
+
   # protected
 
   # If you have extra params to permit, append them to the sanitizer.
   # def configure_sign_in_params
   #   devise_parameter_sanitizer.permit(:sign_in, keys: [:attribute])
   # end
+
+  private
+
+  def check_signed_in
+    return unless signed_in?
+
+    redirect_to after_sign_in_path_for(resource)
+  end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -26,10 +26,46 @@ module ApplicationHelper
   end
 
   def navigation
-    govuk_header(service_name: t("service.name")) do |header|
+    govuk_header(service_name:) do |header|
       case current_namespace
       when "manage"
         if current_staff # TODO: replace with case worker user type
+          header.navigation_item(
+            active: current_page?(main_app.manage_interface_referrals_path),
+            href: main_app.manage_interface_referrals_path,
+            text: "Referrals"
+          )
+          if current_staff.view_support?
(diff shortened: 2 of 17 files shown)


app/controllers/staff/sessions_controller.rb
```diff
@@ -2,6 +2,10 @@
 
 class Staff::SessionsController < Devise::SessionsController
   # before_action :configure_sign_in_params, only: [:create]
+  # rubocop:disable Rails/LexicallyScopedActionFilter
```
paulodeon, line 5: Could this be moved into `.rubocop.yml`?

```ruby
Rails/LexicallyScopedActionFilter:
  Exclude:
    - app/controllers/staff/**/*
```

app/controllers/staff/sessions_controller.rb
```diff
@@ -18,10 +22,36 @@ class Staff::SessionsController < Devise::SessionsController
...
   # protected
 
   # If you have extra params to permit, append them to the sanitizer.
   # def configure_sign_in_params
   #   devise_parameter_sanitizer.permit(:sign_in, keys: [:attribute])
   # end
+
+  private
+
+  def check_signed_in
```
paulodeon, line 45: Can this use `after_sign_in_path_for`? i.e. `redirect_to after_sign_in_path_for`?

spec/support/system/authorization_steps.rb
```diff
@@ -1,26 +1,39 @@
...
+  def when_i_login_as_a_case_worker_with_management_permissions_only
+    create(:staff, :confirmed, :can_manage_referrals)
 
-    sign_in(user)
-  end
+    visit new_staff_session_path
 
-  def when_i_am_authorized_as_a_case_worker_without_management_permissions
-    user = create(:staff, :confirmed, :can_view_support)
+    fill_in "Email", with: "test@example.org"
```
paulodeon, line 7: Any reason not to use the `sign_in` helper here?

app/controllers/staff/sessions_controller.rb
tavy88: We could do that yes.

app/controllers/staff/sessions_controller.rb
tavy88: This is the devise method that I am overriding.

spec/support/system/authorization_steps.rb
tavy88: Yes, I want to get to use the after_sign_in method. `sign_in` will skip that behaviour and just put the user on the session.

app/controllers/staff/sessions_controller.rb
paulodeon: Just seems like to avoid duplication on the url if else we could write the method like this:

```ruby
def check_signed_in
  return unless signed_in?
  redirect_to after_sign_in_path_for(resource)
end
```

tavy88: You're right.

paulodeon (approved):

gpeng (approved):

app/helpers/application_helper.rb
```diff
@@ -26,10 +26,46 @@ def current_namespace
   end
 
   def navigation
-    govuk_header(service_name: t("service.name")) do |header|
+    govuk_header(service_name:) do |header|
       case current_namespace
       when "manage"
         if current_staff # TODO: replace with case worker user type
+          header.navigation_item(
+            active: current_page?(main_app.manage_interface_referrals_path),
```
vassyz, line 34: You already merged it, but the way this menu is built makes it difficult to maintain.

PR #525, opened 2023-02-27
1237 - Use a custom sign in route for case workers

tavy88:
### Context

Use a custom sign in route for case workers

Instead of using `/staff/sign_in` users should be able to use `manage/sign-in` as route.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -13,6 +13,7 @@ Rails.application.routes.draw do
   )
   devise_scope :staff do
     get "/staff/sign_out", to: "staff/sessions#destroy"
+    get "/manage/sign-in", to: "staff/sessions#new"
   end
 
   devise_for(:user)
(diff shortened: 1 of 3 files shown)


config/routes.rb
```diff
@@ -13,6 +13,7 @@
   )
   devise_scope :staff do
     get "/staff/sign_out", to: "staff/sessions#destroy"
+    get "/manage/sign-in", to: "staff/sessions#new"
```
vassyz, line 16: `sign_out`, `sign-in` should we keep the format consistent?

tavy88: I mean we could but it wasn't in the scope of this PR.

vassyz (approved):

gpeng (approved):

PR #526, opened 2023-02-28
1250 - Case workers return to original url after login

tavy88:
### Context

Case workers return to original url after login

If a case worker accesses any url under `/support/*` or `/manage/*`, but they are not authenticated, they will return to the original url after login.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/staff/sessions_controller.rb
+++ b/app/controllers/staff/sessions_controller.rb
@@ -20,7 +20,13 @@ class Staff::SessionsController < Devise::SessionsController
   #   super
   # end
 
-  def after_sign_in_path_for(_resource)
+  def after_sign_in_path_for(resource)
+    stored_location = stored_location_for(resource)
+
+    if stored_location.present? && stored_location.exclude?(manage_sign_in_path)
+      return URI.parse(stored_location).path
+    end
+
     if current_staff.manage_referrals?
       manage_interface_referrals_path
     elsif current_staff.view_support?
@@ -43,6 +49,6 @@ class Staff::SessionsController < Devise::SessionsController
   def check_signed_in
     return unless signed_in?
 
-    redirect_to after_sign_in_path_for(resource)
+    redirect_to after_sign_in_path_for(current_staff)
   end
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -9,6 +9,11 @@ Rails.application.routes.draw do
       passwords: "staff/passwords",
       sessions: "staff/sessions",
       unlocks: "staff/unlocks"
+    },
+    path: "",
+    path_names: {
+      sign_in:  "manage/sign-in",
+      sign_out: "manage/sign-out",
     }
   )
   devise_scope :staff do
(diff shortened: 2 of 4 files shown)


app/controllers/staff/sessions_controller.rb
```diff
@@ -21,6 +21,10 @@ class Staff::SessionsController < Devise::SessionsController
   # end
 
   def after_sign_in_path_for(_resource)
+    path = stored_location || request.referer
+
+    return path if path.present? && path.exclude?(manage_sign_in_path)
+
```
tavy88, line 27: Capybara does not follow redirects so it's pretty hard to test this behaviour here.

paulodeon: Pretty certain capybara does follow redirects and the above should work, have you tried it yet?

tavy88: Yep it doesn' t work :)

app/controllers/staff/sessions_controller.rb
```diff
@@ -21,6 +21,10 @@ class Staff::SessionsController < Devise::SessionsController
   # end
 
   def after_sign_in_path_for(_resource)
+    path = stored_location || request.referer
```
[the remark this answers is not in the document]
tavy88: This exact thing happened to me. I guess it's when you don't have a user in the session in which case it is nil.

config/routes.rb
```diff
@@ -9,6 +9,11 @@
       passwords: "staff/passwords",
       sessions: "staff/sessions",
       unlocks: "staff/unlocks"
+    },
+    path: "",
```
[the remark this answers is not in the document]
tavy88: Yep. Confirmed with @gpeng -> we don't want `/staff/` at all in the urls.

app/controllers/staff/sessions_controller.rb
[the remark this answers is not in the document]
paulodeon: If it's good enough for devise... https://github.com/heartcombo/devise/blob/8593801130f2df94a50863b5db535c272b00efe1/lib/devise/controllers/helpers.rb#L215

gpeng (changes requested):

app/controllers/staff/sessions_controller.rb
```diff
@@ -20,7 +20,13 @@ class Staff::SessionsController < Devise::SessionsController
   #   super
   # end
 
-  def after_sign_in_path_for(_resource)
+  def after_sign_in_path_for(resource)
+    path = stored_location_for(resource)
```
gpeng, line 24: Naming this `path` is a bit confusing if we don't know whether it's a URL or a path. `stored_location` maybe?

spec/system/manage/case_worker_returns_to_original_url_after_login_spec.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+RSpec.feature "Manage referrals and staff" do
+  include CommonSteps
+
+  scenario "Case worker returns to the original path after login",
+           type: :system do
```
gpeng, line 8: I don't think we need type as have infer from location on

gpeng (approved):

PR #527, opened 2023-02-28
TEST ONLY Case workers return to original url after login

tavy88:
If a case worker accesses any url under /support/* or /manage/*, but they are not authenticated,
they will return to the original url after login.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


tavy88:
Not needed anymore.

PR #528, opened 2023-02-28
[1248] Add TRA logo to the referral PDF

gpeng:
### Context

Case workers can download a PDF version of a referral. This PDF is sent to the person the allegation pertains to. It needs to be TRA branded.

* Add the TRA logo

Screenshot 2023-02-28 at 17 11 34

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


app/views/referrals/pdf.html.erb
```diff
@@ -5,9 +5,15 @@
...
         }
+        .logo {
+          margin: 0px 0px 80px 0px;
+          width: 150px;
+          height: 150px;
+        }
     </style>
   </head>
   <body>
+    <img src="<%= @logo %>" class="logo"/>
```
tavy88, line 16: How do you feel about using the rails helper:

```ruby
image_tag
```

here?

tavy88 (approved):

app/views/referrals/pdf.html.erb
gpeng: I did start with that but I think it uses the asset path which doesn't work in this view. Not sure if we can work around that but I decided it was just as easy to build the URL in the job as we're already doing that with the stylesheet.

tavy88: Ah got it.

PR #530, opened 2023-03-01
[1259] Add referrer type to manage view

gpeng:
### Context

We need to explicitly show the source of the referral ie Public/Employer

* Add 'Type: Public/Employer' to the referrer details. This will also appear on the PDF version

### Guidance to review

Screenshot 2023-03-01 at 11 01 49

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/components/manage_interface/referrer_details_component.rb
+++ b/app/components/manage_interface/referrer_details_component.rb
@@ -15,6 +15,7 @@ module ManageInterface
           { key: { text: "Job title" }, value: { text: referrer.job_title } }
         )
       end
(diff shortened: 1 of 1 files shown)


tavy88 (approved):
It would be good for future maybe to add some specs for components as well. What do you think?

gpeng:
> It would be good for future maybe to add some specs for components as well. What do you think?

Yep. Agreed. I've carded up a review for us to discuss and add some https://trello.com/c/tIpooZmN/149-component-tests

PR #532, opened 2023-03-01
Add upload file size hint and tidy file sizes throughout

paulodeon:
### Context

To make it clearer that there is a filesize limit before the user uploads

Show a hint to the users with all file upload fields. The text needs to say "50MB" not "50 MB" as returned by the rails helper so had to create our own. 

Took the liberty of tidying up the file sizes throughout by adding a `FileSizeHelper`

Also fixed a bug where uploading a file on top of an existing file ignored file size validation

### Link to Trello card

[](https://trello.com/c/yLd8A4Ti/1243-referrals-design-history-change-adding-hint-text-to-file-uploads)

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/forms/referrals/allegation/details_form.rb
+++ b/app/forms/referrals/allegation/details_form.rb
@@ -10,11 +10,13 @@ module Referrals
                 if: -> { allegation_format == "details" }
       validates :allegation_upload,
                 presence: true,
-                file_upload: true,
                 if: -> {
                   allegation_format == "upload" &&
                     !referral.allegation_upload.attached?
                 }
+      validates :allegation_upload,
+                file_upload: true,
+                if: -> { allegation_format == "upload" }
 
       attr_accessor :referral,
                     :allegation_details,
--- a/app/helpers/file_size_helper.rb
+++ b/app/helpers/file_size_helper.rb
@@ -1,0 +1,13 @@
+module FileSizeHelper
+  def file_size(attachment)
+    ActiveSupport::NumberHelper.number_to_human_size(
+      attachment.byte_size
+    ).delete(" ")
+  end
+
+  def max_allowed_file_size
+    ActiveSupport::NumberHelper.number_to_human_size(
+      FileUploadValidator::MAX_FILE_SIZE
+    ).delete(" ")
+  end
+end
--- a/spec/helpers/file_size_helper_spec.rb
+++ b/spec/helpers/file_size_helper_spec.rb
@@ -1,0 +1,23 @@
+require "rails_helper"
+
+RSpec.describe FileSizeHelper, type: :helper do
+  describe "#file_size" do
+    subject { file_size(attachment) }
+
+    context "when called with an attachment" do
+      let(:attachment) do
+        instance_double("ActiveStorage::Blob", byte_size: 33.6.kilobytes)
+      end
+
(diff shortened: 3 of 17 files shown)


app/forms/referrals/allegation/details_form.rb
```diff
@@ -10,11 +10,13 @@ class DetailsForm
...
       validates :allegation_upload,
                 presence: true,
-                file_upload: true,
                 if: -> {
                   allegation_format == "upload" &&
                     !referral.allegation_upload.attached?
                 }
+      validates :allegation_upload,
+                file_upload: true,
+                if: -> { allegation_format == "upload" }
```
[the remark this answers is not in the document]
paulodeon: The conditionals are different for each

We dont want to validate `presence` when there is a file attached, or we get an validation error when we try and save a form like `/referrals/58/previous-misconduct/detailed-account/edit` when we've already uploaded a file and are not uploading a new one.

The `file_upload` validation only runs on files actually uploaded. We do want to run that validation when we save the above form as we could be uploading a file at that point.

There is an argument for factoring presence into the file upload validator but that's a bigger change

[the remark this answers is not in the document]
tavy88: Ok makes sense. Now I know the rationale behind it.

app/helpers/file_size_helper.rb
```diff
@@ -0,0 +1,15 @@
...
+      .join
+  end
+
+  def max_allowed_file_size
+    ActiveSupport::NumberHelper
+      .number_to_human_size(FileUploadValidator::MAX_FILE_SIZE)
+      .split(" ")
+      .join
+  end
+end
```
tavy88, line 15: Unit tests for this helper?

paulodeon: Given it's simplicity and relative lack of importance I'm not sure that's strictly necessary here

vassyz: We haven't tested the helpers individually, but I tend to agree with @tavy88. Having a spec to quickly output the result wouldn't hurt.

vassyz: Do we really need to remove the spaces? I'm leaning towards using the default format.

paulodeon: @adamsilver was fairly firm on that. I'll add a quick test for it though

vassyz: Don't really agree with it, but not worth debating to much about it. Why not something like `.gsub(' ', '')` or `.delete (' ')` though?

paulodeon: Added a test

spec/helpers/file_size_helper_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  describe "#max_allowed_file_size" do
+    subject { max_allowed_file_size }
+
+    context "when called returns the max allowed file size human formatted with space removed" do
+      it { is_expected.to eq "50MB" }
+    end
+  end
+end
```
tavy88, line 23: Amazing!

tavy88 (approved):
LGTM

vassyz (approved):

PR #535, opened 2023-03-01
Edit staff user roles

vassyz:
### Context

Add edit user form with roles checkboxes in the support interface. Accessible by view_support permission on the invite and "Edit permissions" pages.

- the invite user form has two checkboxes ‘Permissions’ with options ‘view referrals’ and ‘view support’
- at least one is required (validate with an error summary)
- the user list has an ‘change permissions' link (under the email) that opens an edit form with the checkboxes

Screenshot 2023-03-01 at 09 52 11
Screenshot 2023-03-01 at 09 52 17
Screenshot 2023-03-01 at 09 53 48
Screenshot 2023-03-01 at 09 53 55
Screenshot 2023-03-01 at 09 54 01

### Guidance to review

I couldn't use the same logic for both forms because one uses devise and the other uses a form object. This causes some duplication related to validation.

### Link to Trello card

https://trello.com/c/Bkdnrj3D/1222-edit-staff-user-roles

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


vassyz:
[Edit staff user roles](https://trello.com/c/Bkdnrj3D/1222-edit-staff-user-roles)

app/models/staff.rb
```diff
@@ -26,4 +28,10 @@ def password_complexity
   def send_devise_notification(notification, *args)
     devise_mailer.send(notification, self, *args).deliver_later
   end
+
+  def permissions_are_valid
+    return if manage_referrals? || view_support?
+
+    errors.add(:permissions, "Select permissions")
```
tavy88, line 35: Not a big issue now, but I think this message should live in the `locales.yml` file.

vassyz: I couldn't get a translation scope to work so I'd have to give it the full path, which is a bit ugly. I'll probably end up doing that but was hoping somebody has a solution.

app/controllers/support_interface/staff_permissions_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+          manage_referrals: @staff.manage_referrals
+        )
+    end
+
+    def update
+      @staff_permissions_form =
+        StaffPermissionsForm.new(staff_params.merge(staff: @staff))
+
+      if @staff_permissions_form.save
+        flash[:success] = "Staff permissions updated for #{@staff.email}"
```
tavy88, line 18: Same here. It would be better to have this message in the `locales.yml`.

app/controllers/support_interface/staff_permissions_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    def staff_params
+      params.require(:support_interface_staff_permissions_form).permit(
+        :manage_referrals,
+        :view_support
+      )
+    end
+
+    def set_staff
+      @staff = Staff.find(params[:id])
+    end
```
tavy88, line 36: Does this need to be a `before_action`?

You can just do something like:

``` ruby
def staff_user
   @staff_user ||= Staff.find(params[:id])
end
```

and then you can just use `staff_user` in any controller actions you want.

app/forms/support_interface/staff_permissions_form.rb
```diff
@@ -0,0 +1,24 @@
...
+
+      staff.update(view_support:, manage_referrals:)
+    end
+
+    private
+
+    def permissions_are_valid
+      return if manage_referrals || view_support
+
+      errors.add(:permissions, "Select permissions")
```
tavy88, line 21: It would be good to have this message in the locales.yml.

app/models/staff.rb
tavy88: That's interesting. Let's jump on a call.

tavy88 (approved):

PR #539, opened 2023-03-02
1254 - Manage referrals - Teacher Personal Details

tavy88:
Add missing fields on the form visible inside the
manage referrals - Personal details section.

### Context

Add missing fields on the form visible inside the
manage referrals - Personal details section.

Add missing fields on the form visible on manage referrals - personal details section

##### Public referral

Screenshot 2023-03-02 at 16 08 03

##### Employer Referral

Screenshot 2023-03-02 at 16 08 16

##### Public referral

Screenshot 2023-03-02 at 16 09 37

##### Employer referral

Screenshot 2023-03-02 at 16 09 25

### Link to Trello card

https://trello.com/c/xGHLYcN3

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/components/manage_interface/personal_details_component.rb
+++ b/app/components/manage_interface/personal_details_component.rb
@@ -9,7 +9,80 @@ module ManageInterface
         { key: { text: "First name" }, value: { text: referral.first_name } },
         { key: { text: "Last name" }, value: { text: referral.last_name } }
       ]
+
+      rows.push(
+        {
+          key: {
+            text: "Are they known by other name?"
+          },
+          value: {
+            text: referral.name_has_changed
+          }
+        }
+      )
+
+      if referral.previous_name.present?
+        rows.push(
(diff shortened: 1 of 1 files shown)


gpeng (approved):
Still not sure we need the 'yes' bits when we have the subsequent value as we discussed but I think the most important bit of this is to make sure first we have all of the data and then we can refine what the user sees based on that. So 👍 from me but I'm interested to hear if @adamsilver & @Dan13L3 agree. Even if they don't I'd be keen to merge this and refine in new PRs given the time pressure.

tavy88:
Agreed!

vassyz:
Shouldn't we use title case? `"Yes" : "No"` To keep it consistent with the other components.

tavy88:
ok we can do!

PR #546, opened 2023-03-03
Validate dates before 1920

vassyz:
### Context
Added to date of birth, job start date, job end date pages.

Screenshot 2023-03-03 at 14 14 17

### Link to Trello card

https://trello.com/c/uZAhKWwY/1231-employer-form-validate-dates-before-1920

### DIFF
--- a/app/validators/date_validator.rb
+++ b/app/validators/date_validator.rb
@@ -78,27 +78,25 @@ class DateValidator < ActiveModel::EachValidator
       return false
     end
 
-    return true unless options[:date_of_birth] || !options[:in_the_future]
-
-    if year > Time.zone.today.year
-      record.errors.add(attribute, :in_the_future)
+    if options[:not_future] && Date.new(year, month, day) > Time.zone.today
+      record.errors.add(attribute, :not_future)
       return false
     end
 
-    return true unless options[:date_of_birth]
-
-    if year < 1900
-      record.errors.add(attribute, :born_after_1900)
+    if options[:past_century] && !year.between?(1920, Time.zone.today.year)
+      record.errors.add(attribute, :past_century)
       return false
     end
 
-    if record.send(attribute) > 16.years.ago
-      record.errors.add(attribute, :inclusion)
+    if options[:above_16] && record.send(attribute) > 16.years.ago
+      record.errors.add(attribute, :above_16)
       return false
     end
 
     true
   end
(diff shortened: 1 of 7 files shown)


app/validators/date_validator.rb
```diff
@@ -78,22 +78,19 @@ def validate_each(record, attribute, _value)
       return false
     end
 
-    return true unless options[:date_of_birth] || !options[:in_the_future]
-
-    if year > Time.zone.today.year
+    if (options[:past_century] || !options[:in_the_future]) &&
```
paulodeon, line 81: Seems a bit confusing to mix up the two validation options here, can that be avoided?

Also it seems that `:in_the_future` actually means in the past which is pretty confusing!

app/validators/date_validator.rb
```diff
@@ -78,22 +78,19 @@ def validate_each(record, attribute, _value)
...
       record.errors.add(attribute, :in_the_future)
       return false
     end
 
-    return true unless options[:date_of_birth]
-
-    if year < 1900
-      record.errors.add(attribute, :born_after_1900)
+    if options[:past_century] && year < 1920
+      record.errors.add(attribute, :before_century)
```
paulodeon, line 88: I would prefer if the keys for validation and translation matched here

config/locales/en.yml
```diff
@@ -84,8 +84,8 @@ en:
           attributes:
             date_of_birth:
               blank: Enter their date of birth
-              born_after_1900: Enter a year later than 1900
-              inclusion: You must be 16 or over to use this service
+              before_century: Their date of birth must be the same as or after 1 January 1920
```
paulodeon, line 87: This seems a little wordy, worth getting Daniel or Adam to check?

spec/validators/date_validator_spec.rb
```diff
@@ -97,7 +121,7 @@ def stub_validatable_class(name, options: true)
       it { is_expected.to be_invalid }
     end
 
-    context "with max age date of birth params" do
+    context "with year fefore 1920" do
```
paulodeon, line 124: Spelling

config/locales/en.yml
vassyz: Text came from @adamsilver

paulodeon (approved):

PR #548, opened 2023-03-03
Remove questions that have yes as answer

tavy88:
### Context

Remove questions that have yes as answer

If we have conditional questions with their answer Yes we don't show them.

### Link to Trello card

https://trello.com/c/xGHLYcN3

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/components/manage_interface/employment_details_component.rb
+++ b/app/components/manage_interface/employment_details_component.rb
@@ -35,59 +35,63 @@ module ManageInterface
       ]
 
       if referral.from_member_of_public?
-        rows.push(
-          {
-            key: {
-              text:
-                "Is name and address of the organisation where the alleged misconduct took place known?"
-            },
-            value: {
-              text: referral.organisation_address_known ? "Yes" : "No"
+        if !referral.organisation_address_known
+          rows.push(
+            {
+              key: {
+                text:
+                  "Do you know the name and address of the organisation where the alleged misconduct took place?"
+              },
+              value: {
+                text: "No"
+              }
             }
-          }
-        )
-
-        rows.push(
-          {
-            key: {
-              text:
-                "Name and address of the organisation where the alleged misconduct took place"
-            },
-            value: {
-              text: referral_organisation_address(referral)
+          )
+        else
+          rows.push(
+            {
+              key: {
+                text:
+                  "Name and address of the organisation where the alleged misconduct took place"
+              },
+              value: {
+                text: referral_organisation_address(referral)
+              }
             }
-          }
-        )
+          )
+        end
       end
 
       return rows unless referral.from_employer?
 
-      rows.push(
-        {
-          key: {
-            text:
-              "Were they employed at the same organisation as you at the time of the alleged misconduct?"
-          },
-          value: {
-            text: referral.same_organisation ? "Yes" : "No"
+      if !referral.same_organisation
+        rows.push(
+          {
+            key: {
+              text:
+                "Were they employed at the same organisation as you at the time of the alleged misconduct?"
+            },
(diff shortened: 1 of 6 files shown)


app/components/manage_interface/employment_details_component.rb
```diff
@@ -135,41 +139,43 @@ def rows
...
-        )
-
-        rows.push(
-          {
-            key: {
-              text: "Are they employed somewhere else?"
-            },
-            value: {
-              text: working_somewhere_else
+          rows.push(
```
vassyz, line 143: Was putting this in the `else` branch of `referral.role_end_date_known` a mistake?

app/components/manage_interface/employment_details_component.rb
```diff
@@ -135,41 +139,43 @@ def rows
...
+              {
+                key: {
+                  text: "Are they employed somewhere else?"
+                },
+                value: {
+                  text: working_somewhere_else
+                }
+              }
+            )
+          elsif !referral.work_location_known
```
vassyz, line 165: I think this was missing from the original page as well, but should't it display the `organisation_name` in another `else` branch here?

app/components/manage_interface/employment_details_component.rb
tavy88: Not really because I think I got to these questions only if the person has left the job.

app/components/manage_interface/employment_details_component.rb
tavy88: There already is a section called `Other Employment` where this missing bit is added.

app/components/manage_interface/employment_details_component.rb
vassyz: Yes, but is't that the `if employment_status(referral) == "No"` `else` branch?

app/components/manage_interface/employment_details_component.rb
vassyz: Sorry, missed that, wasn't obvious from the code in that file. 👍

app/components/manage_interface/employment_details_component.rb
tavy88: Uhm no I think it needs to be under the `referral.role_end_date_known` else branch because we start asking them about `Job end date` only if there is a `referral.role_end_date_known` present.

tavy88: Ah I see what you mean. I think you're right.

vassyz:
```ruby
if !referral.role_end_date_known
  rows.push(
    {
      key: {
        text: "Do you know when they left the job?"
      },
      value: {
        text: "No"
      }
    }
  )
else
  rows.push(
    {
      key: {
        text: "Job end date"
      },
      value: {
        text: referral.role_end_date&.to_fs(:long_ordinal_uk)
      }
    }
  )
end
```

This is what I mean, that `end` at the end, so it doesn't contain all the other rows. `reason_leaving_role`, `work_location_known` etc shouldn't be in the `if !referral.role_end_date_known` else branch.

vassyz: To make it easier to follow, you can create methods for all rows, like `def organisation_row`

tavy88: We will need to refactor the whole components already spoke with @gpeng about this but we will do this as a different ticket.

vassyz (approved):
Looks good, just the liner fix.

gpeng (approved):

PR #551, opened 2023-03-05
[1279] Remove old start page

gpeng:
### Context

We are removing the in-service start page as it should live on GOV.UK. The GOV.UK version will be published on 6/3/2023.

* Remove the start page
* Use the `PagesController#start` action to redirect appropriately as there is quite a bit of logic about what page should be shown from the root_path
* Refactor the path logic into `StartPath`

There is still some 'redirect' logic left in `ApplicationController` as this is used by the create referral action. The behaviour of that is strange and we should refactor that too but it's out of scope of this PR.

Moving the start page broke a lot of system specs so I've committed those separately from the removal. We can remove a couple of feature flags this week after launch and then review the system specs as the feature flags are causing them to be overly complicated.

### Guidance to review

Visiting the service domain root should generally take the user to the 'Do you have an account' page. If they are already logged in the behaviour changes depending on whether they have a referral and whether the at is partially completed or submitted.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


app/lib/start_path.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    referral_type_path
+  end
+
+  def referrals_submitted?
+    user.referrals.submitted.any?
+  end
+
+  def edit_path_for_referral(referral)
+    if referral.from_member_of_public?
```
vassyz, line 44: This can be `polymorphic_path([:edit, referral.routing_scope, referral])`

app/lib/start_path.rb
```diff
@@ -0,0 +1,50 @@
+class StartPath
+  include Rails.application.routes.url_helpers
+
+  class << self
```
vassyz, line 4: Does this have a purpose other than not having to call `StartPath.new(user:).start_path`? And not sure how I feel about having this class in `lib`. Why not remove it from pages and create something like a `StartController` or `RootController`?

gpeng: It just makes it simpler to call and enforces the interface. Even if we had a separate controller for this action I'd want to extract this stuff into a class as I think there is too much logic to test in a controller.

dankmitchell (approved):

PR #556, opened 2023-03-06
Fix work location address visibility

paulodeon:
### Context

Work location address visibility was wrong in manage and on the PDF

What I found is that the address did show correctly if it is entered. But it would sometimes show as blank if the no radio was checked on work location known so I fixed that

### Link to Trello card

[](https://trello.com/c/B37qRbzo/1281-other-employer-details)

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/views/manage_interface/referrals/show.html.erb
+++ b/app/views/manage_interface/referrals/show.html.erb
@@ -22,12 +22,12 @@
     <%= render ManageInterface::PersonalDetailsComponent.new(referral: @referral) %>
     <%= render ManageInterface::EmploymentDetailsComponent.new(referral: @referral) %>
 
-    <% if @referral.from_employer? %>
+    <% if @referral.work_location_known? %>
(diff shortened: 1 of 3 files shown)


app/views/manage_interface/referrals/show.html.erb
```diff
@@ -22,12 +22,12 @@
     <%= render ManageInterface::PersonalDetailsComponent.new(referral: @referral) %>
     <%= render ManageInterface::EmploymentDetailsComponent.new(referral: @referral) %>
 
-    <% if @referral.from_employer? %>
+    <% if @referral.from_employer? && @referral.work_location_known? %>
```
vassyz, line 25: I think technically, `if @referral.work_location_known?` is enough.

paulodeon: yep, fixed

vassyz: You only changed it in the pdf.

vassyz (approved):

PR #559, opened 2023-03-06
Send confirmation email

vassyz:
### Context

When a user submits a referral they will receive a confirmation email.

### Link to Trello card

https://trello.com/c/7M27fCB9/1282-send-confirmation-email

### DIFF
--- a/app/views/user_mailer/referral_submitted.erb
+++ b/app/views/user_mailer/referral_submitted.erb
@@ -1,0 +1,24 @@
+You’ve sent a referral of serious misconduct.
+
+View your referral:
+
+<%= @link %>
+
+WHAT HAPPENS NEXT
+
+Your referral will usually be reviewed within 3 working days. It may be longer if more information is needed.
+
+It could be decided that:
+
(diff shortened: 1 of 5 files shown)


app/views/user_mailer/referral_submitted.erb
```diff
@@ -0,0 +1,23 @@
...
+View your referral:
+
+<%= @link %>
+
+WHAT HAPPENS NEXT
+
+Your referral will usually be reviewed within 3 working days. It may be longer if more information is needed.
+
+It could be decided that:
+• the case will not be investigated, for example because it’s unlikely the teacher will be stopped from teaching
```
gpeng, line 12: Are these bullets correct for notify? Do they need to be `*`-ses with a line space above the first one?

gpeng (approved):

app/views/user_mailer/referral_submitted.erb
vassyz: You're right, missed that.

PR #567, opened 2023-03-08
Fix links in referral PDF

paulodeon:
### Context

Links in the PDF weren't working and couldn't work as they are expiring URLs so we replaced them with filenames

### Link to Trello card

https://trello.com/c/OGu8lUE6/1242-fix-links-in-the-referral-pdf

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/spec/jobs/render_pdf_job_spec.rb
+++ b/spec/jobs/render_pdf_job_spec.rb
@@ -3,6 +3,17 @@ require "rails_helper"
 RSpec.describe RenderPdfJob do
   let(:referral) { create(:referral, :complete, :with_attachments) }
 
+  describe "generating the HTML" do
+    let(:job) { described_class.new }
+
+    it "renders the html and removes the a tags" do
+      job.perform(referral:)
+
+      expect(job.send(:html)).to match(%r{<dd.*?><a.*?>upload1.pdf</a></dd>})
+      expect(job.send(:processed_html)).to match(%r{<dd.*?>upload1.pdf</dd>})
+    end
+  end
+
   describe "running the job" do
     subject(:perform) { described_class.new.perform(referral:) }
 
--- a/app/jobs/render_pdf_job.rb
+++ b/app/jobs/render_pdf_job.rb
@@ -28,6 +28,10 @@ class RenderPdfJob < ApplicationJob
     )
   end
 
+  def processed_html
+    html&.gsub(%r{<a.*?>(.*?)</a>}, '\1')
+  end
+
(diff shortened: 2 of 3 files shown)


vassyz:
You could just parse the HTML before it gets sent to Grover. Something like:

```ruby
def io
  StringIO.new(Grover.new(html_without_links, format: "A4").to_pdf)
end

def html_without_links
  html.gsub(/<a.*?>(.*?)<\/a>/, '\1')
end
```

Or use Nokogiri if you prefer.

paulodeon:
That's certainly a simpler solution than what I've proposed. Although I suppose it's possible that in future it may be worth having a different rendering path for the PDF than for the pages. Any thoughts on this @gpeng ? I'm happy to revert to the simpler solution if we don't think we'll need this in future

gpeng:
It would be good if we could avoid coupling the components too tightly with HTML vs PDF. Seb's suggestion is definitely simpler but I don't think it's just stripping the `a` tags altogether. What does the PDF render if we do that? Does it still render the URL or just the filename. If just the filename then probably go with that for now.

vassyz:
That was just an example. I checked if Grover/Puppeteer have any settings for parsing the HTML, but couldn't find something straightforward. You can run a JS script on initialisation, so that might work. That snipped works fine for me, it renders the filename.

paulodeon:
I do think Sebs solution will work well. I'll go ahead and implement it

paulodeon:
OK, that's done now

spec/jobs/render_pdf_job_spec.rb
```diff
@@ -3,6 +3,19 @@
 RSpec.describe RenderPdfJob do
   let(:referral) { create(:referral, :complete, :with_attachments) }
 
+  describe "generating the HTML" do
+    let(:job) { described_class.new }
+
+    it "renders the html and removes the a tags" do
+      job.perform(referral:)
+
+      expect(job.send(:html)).to include('<a class="govuk-link" href=')
```
vassyz, line 12: Should we update the spec so it checks if the `a` tag content is still in the PDF?

paulodeon: Not sure I understand...isn't that what this line is doing?

paulodeon: I updated it with a regex match which is cleaner I think

vassyz (approved):
Looks good 👍

PR #568, opened 2023-03-08
Fix sign out bug

vassyz:
Clicking the sign out link in the manage interface header was not working properly. It was redirecting to the user sign in page instead of the staff sign in page.

Before
Screenshot 2023-03-08 at 15 00 00

After
Screenshot 2023-03-08 at 14 59 52


vassyz:
[Signout link from manage should go to the manage login](https://trello.com/c/gMA52rSx/1305-signout-link-from-manage-should-go-to-the-manage-login)

PR #572, opened 2023-03-09
Update incorrect login attempts error messages

vassyz:
### Context

Instead of redirecting the user to a page, similar to the referrer OTP logic, I have updated the error message to be more descriptive. This is because redirecting the user to a page would require some additional Devise refactoring that I don't think is worth it.

### Link to Trello card

https://trello.com/c/pLiux5lp/1245-referrals-too-many-sign-in-attempts-caseworker


vassyz:
[Referrals - Too many sign in attempts (Caseworker)](https://trello.com/c/pLiux5lp/1245-referrals-too-many-sign-in-attempts-caseworker)

spec/system/manage/case_worker_gets_locked_after_incorrect_login_attempts_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+    FeatureFlags::FeatureFlag.activate(:service_open)
+  end
+
+  def and_the_eligibility_screener_is_enabled
+    FeatureFlags::FeatureFlag.activate(:eligibility_screener)
+  end
+
+  def and_the_referral_form_feature_is_active
+    FeatureFlags::FeatureFlag.activate(:referral_form)
+  end
```
tavy88, line 43: These steps are already in `ActivateFeaturesSteps` module. I've moved them there.

vassyz: 👍  I removed it from a different file and forgot to do it here.

tavy88 (approved):

PR #581, opened 2023-03-13
Store declaration content on submit

paulodeon:
### Context

The declaration could change and we would currently have to review commits and deploy times to figure out which users agreed to which declaration. 

Separate the declaration into a partial
Render a cleaned version of that partial into a db field on the referral

### Guidance to review

An alternative solution which I did do a spike on is to save the version of the partial - this would have to rely on an ENV var to set the current version which would be vulnerable to configuration error and have problems with rollbacks.

The proposed solution seems the cleanest way to achieve this.

### Link to Trello card

https://trello.com/c/p7FqEU5w/1303-store-the-declaration-content-that-the-user-has-agreed-to-with-the-referral

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/lib/declaration_renderer.rb
+++ b/app/lib/declaration_renderer.rb
@@ -1,0 +1,15 @@
+class DeclarationRenderer
+  include ActionView::Helpers::SanitizeHelper
+
+  def render
+    strip_tags(rendered_declaration_partial)
+      .lines
+      .reject(&:blank?)
+      .each(&:strip!)
+      .join("\n")
+  end
+
+  def rendered_declaration_partial
+    ApplicationController.renderer.new.render(partial: "shared/declaration")
+  end
+end
--- a/app/models/referral.rb
+++ b/app/models/referral.rb
@@ -37,6 +37,8 @@ class Referral < ApplicationRecord
   delegate :name, to: :referrer, prefix: true, allow_nil: true
 
   def submit
+    self.declaration = DeclarationRenderer.new.render
+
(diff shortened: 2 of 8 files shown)


gpeng (approved):
This looks fine. I know it's a lot of denormalised stuff but this is going to be a small table for a long time (famous last words) and it meets the need we have for now. It will be simple enough as an iteration to normalise it and add admin tools to update if we have a need in the future.

We should back-fill the existing records too though. I've a preference for not doing that in schema migrations but open to suggestions. Data migrations? Given there are < 20 existing records and they're all the same we could do it in the console in this case.

vassyz (approved):
I wouldn't have used an ENV var, but a database table similar to migrations. It saves a version and a boolean if it's the current one. Then you never update the partial but add a different version to it e.g. `_declaration_v1`. You link this to the submitted referral and on the website you display the current one. With this approach you do need a way of updating the database, rake task, migration or manually. But I think at least you can easily see how many versions we've had and who agreed to what instead of some text.

paulodeon:
Here's the snippet we can run on the prod console: 
```ruby
Referral.where.not(submitted_at: nil).each { |r| r.update(declaration: DeclarationRenderer.new.render) }
```
Agree the db migration idea is better than an ENV var but still it adds some complexity. I guess let's see how needs evolve. I don't think it's too hard to switch over to a versioning system later, but I'm also guessing it's unlikely that will become a requirement.

PR #582, opened 2023-03-13
1301 - Allow staff users to be soft deleted

tavy88:
### Context

Caseworkers can now soft delete other users

We decided to soft delete users for auditing purposes.

- allow staff users to soft delete other staff users
- do not show deleted staff users in the UI

Screenshot 2023-03-15 at 08 50 35

Screenshot 2023-03-15 at 09 09 38

### Link to Trello card

https://trello.com/c/ZCQHoPWy

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -18,6 +18,7 @@ Rails/ApplicationController:
 Rails/LexicallyScopedActionFilter:
   Exclude:
     - app/controllers/staff/**/*
+    - app/controllers/support_interface/staff_controller.rb
 
 Rails/SaveBang:
   Enabled: false
--- a/app/controllers/support_interface/staff_controller.rb
+++ b/app/controllers/support_interface/staff_controller.rb
@@ -1,7 +1,35 @@
 module SupportInterface
   class StaffController < SupportInterfaceController
+    before_action :can_delete, only: %i[destroy delete]
+
     def index
-      @staff = Staff.order(:email)
+      @staff = Staff.active.order(:email)
+    end
+
+    def destroy
+      staff_user.archive
+
+      flash[:info] = if staff_user.save
+        "User deleted"
+      else
+        "User could not be deleted"
+      end
+
+      redirect_to support_interface_staff_index_path
+    end
+
+    private
+
+    def can_delete
+      if current_staff.id == staff_user.id
+        flash[:info] = "You cannot delete your own user"
+
+        redirect_to support_interface_staff_index_path and return
+      end
+    end
+
+    def staff_user
+      @staff_user ||= Staff.find(params[:id])
     end
   end
 end
--- a/app/models/staff.rb
+++ b/app/models/staff.rb
@@ -12,8 +12,11 @@ class Staff < ApplicationRecord
     validate_on_invite: true
   )
 
+  validates :email, uniqueness: true
   validate :password_complexity
   validate :permissions_are_valid
+
+  scope :active, -> { where(deleted_at: nil) }
 
   def password_complexity
     if password.blank? ||
@@ -22,6 +25,10 @@ class Staff < ApplicationRecord
     end
 
     errors.add(:password, :password_complexity)
+  end
+
+  def archive
+    self.deleted_at = Time.zone.now
   end
 
   def send_devise_notification(notification, *args)
@@ -35,5 +42,15 @@ class Staff < ApplicationRecord
       :permissions,
       I18n.t("validation_errors.missing_staff_permission")
     )
+  end
+
+  def active_for_authentication?
+    super && active?
+  end
+
+  private
+
+  def active?
+    deleted_at.nil?
   end
 end
--- a/spec/models/staff_spec.rb
+++ b/spec/models/staff_spec.rb
@@ -42,10 +42,28 @@ RSpec.describe Staff, type: :model do
       end
     end
 
+    context "when the email address is already taken" do
+      before { create(:staff) }
+
+      let(:staff) { build(:staff) }
+
+      it { is_expected.to be_falsey }
+    end
+
     context "when the permissions are invalid" do
       let(:staff) { build(:staff, view_support: nil, manage_referrals: nil) }
 
       it { is_expected.to be_falsey }
+    end
+  end
+
+  describe "#archive" do
+    let(:staff) { create(:staff) }
+
+    it "marks the user as deleted" do
+      staff.archive
+
+      expect(staff.deleted_at).to be_within(1).of(Time.zone.now)
     end
   end
 end
--- a/spec/system/manage/case_workers_cannot_self_delete_spec.rb
+++ b/spec/system/manage/case_workers_cannot_self_delete_spec.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+RSpec.feature "Staff actions" do
+  include CommonSteps
+
+  scenario "Staff users with permissions cannot self delete", type: :system do
+    given_the_service_is_open
+    and_the_eligibility_screener_is_enabled
+
+    when_i_login_as_a_case_worker_with_support_permissions_only
+    and_i_try_to_delete_my_user
+    then_i_get_redirected_to_staff_index_page
+    and_i_see_the_notification_message
+    and_i_am_not_marked_as_deleted
+  end
+
+  private
+
+  def and_i_try_to_delete_my_user
+    @current_staff = Staff.last
+
+    visit delete_support_interface_staff_path(@current_staff)
+  end
+
+  def then_i_get_redirected_to_staff_index_page
+    expect(page).to have_current_path(support_interface_staff_index_path)
+  end
+
+  def and_i_see_the_notification_message
+    expect(page).to have_content("You cannot delete your own user")
+  end
+
+  def and_i_am_not_marked_as_deleted
+    expect(@current_staff.deleted_at).to be_nil
+  end
+end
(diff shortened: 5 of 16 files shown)


app/controllers/support_interface/staff_controller.rb
```diff
@@ -1,7 +1,35 @@
...
   class StaffController < SupportInterfaceController
+    before_action :can_delete, only: %i[destroy delete]
+
     def index
-      @staff = Staff.order(:email)
+      @staff = Staff.active.order(:email)
+    end
+
+    def destroy
+      staff_user.deleted_at = Time.zone.now
```
paulodeon, line 10: Can this be put on the model (or in a concern) as something like `user.soft_delete'?

app/controllers/support_interface/staff_controller.rb
```diff
@@ -1,7 +1,35 @@
 module SupportInterface
   class StaffController < SupportInterfaceController
+    before_action :can_delete, only: %i[destroy delete]
```
paulodeon, line 3: Total nitpicking here but can this be done with Pundit?

.rubocop.yml
```diff
@@ -18,6 +18,7 @@ Rails/ApplicationController:
 Rails/LexicallyScopedActionFilter:
   Exclude:
     - app/controllers/staff/**/*
+    - app/controllers/support_interface/staff_controller.rb
```
paulodeon, line 21: Could we just get rid of this cop everywhere?

app/models/staff.rb
```diff
@@ -12,9 +12,12 @@ class Staff < ApplicationRecord
     validate_on_invite: true
   )
 
+  validates :email, uniqueness: true
```
paulodeon, line 15: What happens when we try and create a user with the same email as a deleted user?

app/controllers/support_interface/staff_controller.rb
paulodeon: could also add the method `soft_deleted?`

tavy88: I'm not sure where would we be using `soft_deleted?` right now. I can move the persistence logic.

app/controllers/support_interface/staff_controller.rb
tavy88: Can you give me an example what you have in mind?

.rubocop.yml
tavy88: Team decision I guess.

app/models/staff.rb
```diff
@@ -36,4 +39,8 @@ def permissions_are_valid
       I18n.t("validation_errors.missing_staff_permission")
     )
   end
+
+  def active_for_authentication?
+    super && !deleted_at
```
paulodeon, line 44: `soft_deleted?` could be used here...

.rubocop.yml
paulodeon: Yeah, I guess I was thinking as we've already disabled it in two areas. Wdy think @vassyz ?

app/controllers/support_interface/staff_controller.rb
paulodeon: I actually don't know pundit very well, but I was thinking it might be possible to create an authorization rule for self deleting - up to you if it's worth it

app/models/staff.rb
tavy88: Screenshot 2023-03-14 at 14 38 01

paulodeon: Do we want that? Because there's no way to un-delete right now right? It would then be impossible to create a user with the same email as a deleted user

app/controllers/support_interface/staff_controller.rb
tavy88: We only have access to `current_staff` within Pundit's context. I feel like we would be overcomplicating things.

.rubocop.yml
vassyz: Fine to disable

app/models/staff.rb
tavy88: It's in the scope of the ticket. We don't want to create duplicated users I think.

app/controllers/support_interface/staff_controller.rb
```diff
@@ -1,7 +1,35 @@
...
+      else
+        "User could not be deleted"
+      end
+
+      redirect_to support_interface_staff_index_path
+    end
+
+    private
+
+    def can_delete
```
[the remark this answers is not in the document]
tavy88: I think I mentioned it this morning as well and @gpeng said that if the see links missing for some users they might get confused? I'm flexible.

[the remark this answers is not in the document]
tavy88: Yep I see your point.

paulodeon (approved):

app/controllers/support_interface/staff_controller.rb
vassyz: I kind of agree with @paulodeon on this one, but also understandable that it might be a bit too much for this PR. Ideally, you'd want to have the @user in the policy, but because the authorisation happens in the parent controller you only have `:support` as a resource. You could also use policies as helpers in views instead of the conditional you added. I'd probably consider this if we have to add similar restrictions to the user actions in the future.

spec/models/staff_spec.rb
```diff
@@ -42,10 +42,28 @@
       end
     end
 
+    context "when the email address is already taken" do
+      before { create(:staff) }
```
vassyz, line 46: I'd probably be more explicit here and add the email as well. When reading the spec you shouldn't have to know that emails are mandatory and that the factory has one.

spec/models/staff_spec.rb
```diff
@@ -42,10 +42,28 @@
...
     end
   end
+
+  describe "#soft_delete" do
+    let(:staff) { create(:staff) }
+
+    it "marks the user as deleted" do
+      staff.soft_delete
+
+      expect(staff.deleted_at).to be_within(1).of(Time.zone.now)
```
vassyz, line 66: I usually travel to time

```ruby
travel_to Time.zone.local(2022, 12, 12)
staff.soft_delete

expect(staff.deleted_at).to eq(Time.zone.local(2022, 12, 12))
```

spec/system/manage/deleted_case_workers_can_no_longer_login_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    visit manage_sign_in_path
+
+    fill_in "Email", with: "test@example.org"
+    fill_in "Password", with: "Example123!"
+
+    click_on "Log in"
+  end
+
+  def and_i_see_the_notification_message
+    expect(page).to have_content("Your account is not activated yet.")
```
vassyz, line 33: Isn't this message a bit confusing? Shouldn't we update the devise inactive translation?

vassyz (approved):

app/controllers/support_interface/staff_controller.rb
```diff
@@ -1,7 +1,35 @@
...
   class StaffController < SupportInterfaceController
+    before_action :can_delete, only: %i[destroy delete]
+
     def index
-      @staff = Staff.order(:email)
+      @staff = Staff.active.order(:email)
+    end
+
+    def destroy
+      staff_user.soft_delete
```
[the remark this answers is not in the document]
tavy88: I was actually thinking the same thing ha ha.

spec/models/staff_spec.rb
tavy88: But the factory does use a static email.

spec/system/manage/deleted_case_workers_can_no_longer_login_spec.rb
tavy88: This is the devise's default message I think. Not sure if it's a problem really. Also, it's very unlikely for most people to see this unless for some reason you are a returning case worker that has been previously deleted.

spec/system/manage/case_workers_cannot_self_delete_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  private
+
+  def and_i_try_to_delete_my_user
+    @current_staff = Staff.last
+
+    visit delete_support_interface_staff_path(@current_staff)
+  end
+
+  def and_i_see_the_notification_message
+    expect(page).to have_content("Your account is not activated yet.")
```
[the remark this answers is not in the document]
tavy88: What would the reason be?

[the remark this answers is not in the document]
vassyz: Because you don't need it.

spec/models/staff_spec.rb
vassyz: I know, read my message again. It was just a thought, not required my any means.

spec/system/manage/deleted_case_workers_can_no_longer_login_spec.rb
vassyz: CCing @adamsilver When you log in after your account has been soft deleted is "Your account is not activated yet". I'd change it to "Your account is not active".

spec/system/manage/case_workers_cannot_self_delete_spec.rb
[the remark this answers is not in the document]
tavy88: Ah yeah I realised it's been defined twice.

spec/system/manage/deleted_case_workers_can_no_longer_login_spec.rb
tavy88: I will check with him in the design review session.

PR #584, opened 2023-03-13
Prevent user from updating submitter referrals

vassyz:
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


vassyz:
Better approach by using controller concerns instead of routing constraints in #597

PR #586, opened 2023-03-14
Add Resend invitation function for staff invites

paulodeon:
### Context

To allow resending invites to staff users

### Link to Trello card

https://trello.com/c/taAGeS9j/1300-resend-invite

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/controllers/support_interface/staff_invitations_controller.rb
+++ b/app/controllers/support_interface/staff_invitations_controller.rb
@@ -1,0 +1,28 @@
+module SupportInterface
+  class StaffInvitationsController < SupportInterfaceController
+    def edit
+      @staff = Staff.find(params[:id])
+    end
+
+    def update
+      @staff = Staff.find(params[:id])
+
+      if @staff.invite!(current_staff)
+        flash[:success] = "Invitation sent"
+        redirect_to support_interface_staff_index_path
+      else
+        flash[:warning] = "Invitation failed to send"
+        render :edit
+      end
+    end
+
+    private
+
+    def staff_params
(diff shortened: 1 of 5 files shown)


app/controllers/support_interface/staff_invitations_controller.rb
```diff
@@ -0,0 +1,31 @@
+module SupportInterface
+  class StaffInvitationsController < SupportInterfaceController
+    before_action :staff
```
tavy88, line 3: I would mention here:

```ruby
before_action :staff, only: %i[edit]
```

because for `update` you end up calling `staff` twice:

```
Once, in the before action
Twice, you are already accessing staff inside update
```

I'm not a fan however, of this pattern personally.

I feel like it's fine to do 

```ruby
 def edit
    @staff = Staff.find(params[:id])
 end
```

and remove the `before_action` completely.

app/controllers/support_interface/staff_invitations_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    def edit
+    end
```
tavy88, line 6: If edit has not controller level logic, I think it's fine to get rid of it completely, rather than keeping it as an empty action.

paulodeon: Personally I like to leave the method there to make it clear what actions are present

paulodeon: Also without `before_action` it is needed now

tavy88 (approved):
LGTM

PR #587, opened 2023-03-14
Set Notify API key before sending every email

vassyz:
### Context

There is still a bug where occasionally email can get sent with an incorrect API key.

Makes sure view_mail never gets called with an incorrect API key. I'm not 100% happy with this solution. I tried creating a general mailer but I struggled with getting the Devise email working. Suggestions welcome.

### Link to Trello card

https://trello.com/c/yumxV5a3/1311-bug-badrequesterror-template-not-found

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -1,28 +1,28 @@
 # frozen_string_literal: true
 class ApplicationMailer < Mail::Notify::Mailer
-  GENERIC_NOTIFY_TEMPLATE = "04b96c89-77ec-4c07-af01-3c96a4b8b5d7"
-
-  MISCONDUCT_GENERIC_NOTIFY_TEMPLATE = "4f8e7de9-3987-4e75-974f-ac94068c4a62"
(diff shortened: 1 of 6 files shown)


app/mailers/application_mailer.rb
```diff
@@ -1,28 +1,28 @@
...
 
-  def set_action_mailer_misconduct_key
+  def set_action_mailer_manage_key
     ActionMailer::Base.notify_settings[:api_key] = ENV.fetch(
       "GOVUK_NOTIFY_MANAGE_SERIOUS_MISCONDUCT_API_KEY"
     )
   end
 
-  def set_action_mailer_default_key
+  def set_action_mailer_tra_key
```
gpeng, line 23: I think this should be `set_action_mailer_refer_key` but we can change that when we switch over to the new service on notify.

gpeng (approved):

app/mailers/application_mailer.rb
vassyz: Done

vassyz:
[BUG: BadRequestError: Template not found](https://trello.com/c/yumxV5a3/1311-bug-badrequesterror-template-not-found)

PR #589, opened 2023-03-14
Reorganise referral/referrer form objects and views

paulodeon:
### Context

Extracted from https://github.com/DFE-Digital/refer-serious-misconduct/pull/517

The referral/referrer forms, controllers and views are in the main view folder unlike other sections

Also instead of a check_answers controller it use a referrers controller

I'm guessing these were built earlier perhaps by someone else

Making routing, naming and file location of forms, controllers, views and check answers consistent for the referrer sections to the rest of the sections.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


tavy88 (approved):
LGTM but check with @vassyz because my understanding was that some of his work might need to go in before this? Also, I know you mentioned on the previous PR that you are confident in unit tests, but from experience moving files around like this sometimes breaks parts of the system that are maybe not that well tested. So I would definitely recommend like a manual happy path test just in case.

paulodeon:
If we are to rename referrals to employer_referrals I'd suggest we do that in a separate PR

paulodeon:
Let me know @vassyz when/if you have work that should be merged in first as it's probably better if I do the merge

gpeng:
> If we are to rename referrals to employer_referrals I'd suggest we do that in a separate PR

Yeah. I agree. This is an improvement as is.

paulodeon:
I have done a happy path test for both the public and employer referral on the review app so I'm feeling pretty comfortable with this.

vassyz:
> Let me know @vassyz when/if you have work that should be merged in first as it's probably better if I do the merge

You can go ahead and merge it.

PR #593, opened 2023-03-16
Increase the line length to 120

vassyz:
### Context

The character limit of 80 was making code harder to read.

Increase the line length to 120 and run the linter.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -24,10 +24,8 @@ Procfile
 Procfile.*
 app/assets/builds
 bin/dev
-config/routes.rb
 data
 db/*.sqlite3
-db/schema.rb
 log
 public/assets
 storage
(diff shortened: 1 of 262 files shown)


.prettierignore
```diff
@@ -24,10 +24,8 @@ Procfile
 Procfile.*
 app/assets/builds
 bin/dev
-config/routes.rb
 data
 db/*.sqlite3
-db/schema.rb
```
paulodeon, line 30: Don't we want to continue to ignore prettier for schema.rb?

vassyz: Why?

paulodeon: As it's an autogenerated file...I don't really mind either way...

paulodeon (approved):

PR #596, opened 2023-03-16
1306 - Use the email templates from the new Notify service

tavy88:
### Context

Move Mailer and Devise Notify templates

Move the existing Mailer and Devise templates from Teaching Regulation Agency
to Refer serious misconduct by a teacher in England service
in Notify.

### Guidance to review

We need to do a full regression testing for any notification that we send to ensure everything is working as expected.

### NOTES

I will need to work with either @vassyz or @gpeng because I don't have Azure access (still) and once the API keys are generated in Notify (you can't see them again) so they have to be added in the vaults as we go.

### Link to Trello card

https://trello.com/c/O36wjl4w

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -1,10 +1,10 @@
 # frozen_string_literal: true
 class ApplicationMailer < Mail::Notify::Mailer
-  TRA_NOTIFY_TEMPLATE = "04b96c89-77ec-4c07-af01-3c96a4b8b5d7"
+  REFER_NOTIFY_TEMPLATE = "8a10a1c1-9cc3-4234-af86-c54cedb7ce09"
   MANAGE_NOTIFY_TEMPLATE = "4f8e7de9-3987-4e75-974f-ac94068c4a62"
   MISCONDUCT_NOTIFY_REPLY_TO_ID = "83cac27e-b914-46a3-a1f9-56e5acb07d05"
 
-  def view_mail_tra(mailer_options:, template_id: TRA_NOTIFY_TEMPLATE)
+  def view_mail_refer(mailer_options:, template_id: REFER_NOTIFY_TEMPLATE)
     set_action_mailer_refer_key
     view_mail(template_id, mailer_options)
   end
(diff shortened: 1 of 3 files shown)


app/mailers/application_mailer.rb
```diff
@@ -1,10 +1,10 @@
 # frozen_string_literal: true
 class ApplicationMailer < Mail::Notify::Mailer
-  TRA_NOTIFY_TEMPLATE = "04b96c89-77ec-4c07-af01-3c96a4b8b5d7"
-  MANAGE_NOTIFY_TEMPLATE = "4f8e7de9-3987-4e75-974f-ac94068c4a62"
-  MISCONDUCT_NOTIFY_REPLY_TO_ID = "83cac27e-b914-46a3-a1f9-56e5acb07d05"
+  MISCONDUCT_NOTIFY_TEMPLATE = "8a10a1c1-9cc3-4234-af86-c54cedb7ce09"
+  MANAGE_NOTIFY_TEMPLATE = "7ef13585-7b41-40ea-a5f7-e8df476b8ea3"
```
gpeng, line 4: We don't need to change the manage templates. We should be using two services in Notify. Emails to referrers should go through the new one we set up but emails to the case workers should come from the existing 'Manage...' service.

gpeng (changes requested):

app/mailers/application_mailer.rb
```diff
@@ -1,10 +1,10 @@
 # frozen_string_literal: true
 class ApplicationMailer < Mail::Notify::Mailer
-  TRA_NOTIFY_TEMPLATE = "04b96c89-77ec-4c07-af01-3c96a4b8b5d7"
+  MISCONDUCT_NOTIFY_TEMPLATE = "8a10a1c1-9cc3-4234-af86-c54cedb7ce09"
```
gpeng, line 3: Sorry to be a pain but the pertinent difference is 'refer' rather than 'manage' so we should name these things with `refer` and `manage rather than misconduct and manage.

vassyz: Sorry, my bad. It should be `refer`.

tavy88: updated

vassyz (approved):

vassyz (approved):

gpeng (approved):

PR #597, opened 2023-03-16
Prevent user from updating submitter referrals

vassyz:
It redirects the user to the referral show page if the referral has been submitted.

### Link to Trello card

https://trello.com/c/4qpIxiR3/1307-stop-submitted-referrals-being-edited-by-the-user

### DIFF
--- a/app/controllers/users/referrals_controller.rb
+++ b/app/controllers/users/referrals_controller.rb
@@ -1,12 +1,19 @@
 module Users
-  class ReferralsController < Referrals::BaseController
+  class ReferralsController < ApplicationController
+    include AuthenticateUser
+    include StoreUserLocation
+
     def index
       @referrals_submitted = current_user.referrals.submitted.order(submitted_at: :desc)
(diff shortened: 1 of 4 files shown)


vassyz:
[Stop submitted referrals being edited by the user](https://trello.com/c/4qpIxiR3/1307-stop-submitted-referrals-being-edited-by-the-user)

app/controllers/users/referrals_controller.rb
```diff
@@ -1,12 +1,16 @@
 module Users
-  class ReferralsController < Referrals::BaseController
+  class ReferralsController < ApplicationController
```
paulodeon, line 2: Wont this open the controller to the public?

vassyz: It wouldn't have been open because you need to have a current user, but it would've crashed the page. I forgot to add the authentication module. Good spot 👍 Fixed now.

paulodeon (approved):

PR #598, opened 2023-03-17
Show file list if present when returning to evidence section

paulodeon:
Previously when uploading a file without completing the evidence section then revisiting that section it would take you to the start of the form.

Now it takes you to the file list if there are files uploaded

See video for demo

https://www.loom.com/share/78299d537900447cabd71bcfb4566e1c

https://trello.com/c/vIMGOLZH/1316-evidence-file-upload-show-list


vassyz (approved):
Looks good. Should probably have a test though.

paulodeon:
Added a test

PR #600, opened 2023-03-20
Delete evidence uploads if has_evidence is set to false

paulodeon:
When a user has uploaded some files but then decides they don’t want to and changes their answer in the check answers pages to “No” we should delete the files they previously uploaded

https://www.loom.com/share/d8199108cd4543fa858d12d67e0d8f25

In addition I have removed the `dependent: :destroy` from all files on the Referral object. It isn't a valid option and was overriding the default `dependent: :purge_later` meaning files weren't being deleted

https://trello.com/c/e8KXLuaT/1317-evidence-file-upload-changing-to-no

### DIFF
--- a/spec/system/public_referrals/user_adds_evidence_spec.rb
+++ b/spec/system/public_referrals/user_adds_evidence_spec.rb
@@ -30,6 +30,22 @@ RSpec.feature "Evidence", type: :system do
     and_i_click_save_and_continue
     then_i_see_a_list_of_the_uploaded_files
 
+    when_i_have_no_more_evidence_to_upload
+    and_i_click_save_and_continue
+    and_i_change_anything_to_upload_to_no
+    and_i_click_save_and_continue
+    and_i_visit_uploaded_files
+    then_the_list_of_the_uploaded_files_should_be_empty
+
+    when_i_have_no_more_evidence_to_upload
+    and_i_click_save_and_continue
+    and_i_change_anything_to_upload_to_yes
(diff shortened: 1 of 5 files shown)


vassyz (approved):
Looks good. There's a mix of `and` and `when` in the specs that doesn't follow our pattern of when-and-then.

tavy88 (approved):

paulodeon:
I fixed that

spec/system/public_referrals/user_adds_evidence_spec.rb
```diff
@@ -30,6 +30,22 @@
...
+    and_i_click_save_and_continue
+    and_i_change_anything_to_upload_to_no
+    and_i_click_save_and_continue
+    and_i_visit_uploaded_files
+    then_the_list_of_the_uploaded_files_should_be_empty
+
+    when_i_have_no_more_evidence_to_upload
+    and_i_click_save_and_continue
+    and_i_change_anything_to_upload_to_yes
+    when_i_choose_yes_to_uploading_evidence
```
vassyz, line 43: Still have it here.

spec/system/public_referrals/user_adds_evidence_spec.rb
```diff
@@ -30,6 +30,22 @@
...
+    and_i_click_save_and_continue
+    when_i_upload_evidence_files
```
vassyz, line 45: And here.

PR #602, opened 2023-03-20
Use Pundit to prevent staff user self-deletion

tavy88:
### Context

Refactor our support policy and use that to prevent staff user self deletion.

Make use of Pundit policies to also prevent staff user self deletion.

### Link to Trello card

https://trello.com/c/W9IgNSc0

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/support_interface/staff_controller.rb
+++ b/app/controllers/support_interface/staff_controller.rb
@@ -1,31 +1,27 @@
 module SupportInterface
   class StaffController < SupportInterfaceController
-    before_action :can_delete, only: %i[destroy delete]
-
     def index
       @staff = Staff.active.order(:email)
     end
 
-    def destroy
-      staff_user.archive
+    def delete
+      authorize staff
+    end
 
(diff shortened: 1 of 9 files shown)


gpeng (approved):

vassyz:
I'd make a few changes to this:

1) Add to `AnonymousSupportUser`:

```ruby
def view_support?
  false
end

def manage_referrals?
  false
end
```
 and remove all the `return false if user.is_a?(AnonymousSupportUser)` lines.

2) Add a new policy

```ruby
class StaffPolicy < ApplicationPolicy
  def delete?
    user.id != record.id
  end

  def destroy?
    user.id != record.id
  end
end
```
and remove the `SupportPolicy` from `StaffController` so it uses the `StaffPolicy`, and the checks from `SupportPolicy` so all it does is `user.view_support?`

```ruby
    def delete
      authorize staff
    end

    def destroy
      authorize staff
```

Let me know if this makes sense.

vassyz (approved):

PR #609, opened 2023-03-22
Add job to delete draft referrals

vassyz:
Deletes the draft referrals older than 90 days, and the associated eligibility checks. This is done in a background job scheduled to run every day at midnight. The main job queues up a separate job for each referral to do the deletion.

### Link to Trello card

https://trello.com/c/XGzaIgIr/1324-automatic-deletion-of-draft-referrals-after-90-days

### DIFF
--- a/app/jobs/delete_draft_referrals_job.rb
+++ b/app/jobs/delete_draft_referrals_job.rb
@@ -1,0 +1,7 @@
+class DeleteDraftReferralsJob < ApplicationJob
+  sidekiq_options queue: "low"
+
+  def perform
+    Referral.stale_drafts.each { |referral| DeleteReferralJob.perform_later(referral.id) }
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -28,6 +28,7 @@ gem "rotp"
 gem "rubyzip"
 gem "sentry-rails"
 gem "sidekiq", "< 7"
+gem "sidekiq-cron"
 gem "tzinfo-data", platforms: %i[mingw mswin x64_mingw jruby]
 gem "uk_postcode"
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -141,6 +141,8 @@ GEM
       railties (>= 3.2)
     e2mmap (0.1.0)
     erubi (1.12.0)
+    et-orbi (1.2.7)
+      tzinfo
     factory_bot (6.2.1)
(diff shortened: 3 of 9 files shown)


app/jobs/delete_draft_referrals_job.rb
```diff
@@ -0,0 +1,13 @@
...
+  sidekiq_options queue: "low"
+
+  def perform
+    draft_referrals.each { |referral| DeleteReferralJob.perform_later(referral.id) }
+  end
+
+  private
+
+  def draft_referrals
+    Referral.draft.where("created_at < ?", 90.days.ago)
```
paulodeon, line 11: Could this be a scope on the model i.e.:

`scope :stale_drafts, -> { where("created_at < ?", 90.days.ago) }`

app/jobs/delete_draft_referrals_job.rb
```diff
@@ -0,0 +1,13 @@
+class DeleteDraftReferralsJob < ApplicationJob
+  sidekiq_options queue: "low"
+
+  def perform
+    draft_referrals.each { |referral| DeleteReferralJob.perform_later(referral.id) }
```
paulodeon, line 5: Is it necessary to create separate jobs for each deletion? Seems a lot of overhead for a destroy? Filesystem deletions would be in their own jobs anyway thanks to `:purge_later`

paulodeon (approved):

app/jobs/delete_draft_referrals_job.rb
paulodeon: It is somewhat nit-picky of me. most likely it makes no difference to sidekiq either way

gpeng: I prefer to do it like this for two reasons. Mostly because it fails better. If one job doesn't work then it is on its own in the retry set to investigate and the others still go through. You can also run the code on one referral easily if you need to in the console or whatever. Admittedly, where it is as simple as this that latter bit isn't so much of a thing and it's very unlikely to fail. Saying that, this will be deleting files on azure so there's a fair possibility of an error in there.

paulodeon: Solid rationale, thanks

vassyz: Thank you both for your input, I'll merge then.

PR #614, opened 2023-03-24
Strip spaces in OTP input

paulodeon:
OTPs are copied and pasted from an email. It may be possible to select whitespace while doing that.

The fix here is to strip the otp param in `user_params`

Devise does have a parameter filter which is configured in the initializer with `strip_whitespace_keys` This applies to email addresses when logging in but not to the OTP as it is applied on `save` not to the params themselves. Hence the slightly hacky approach

https://trello.com/c/jhcJstSI/1315-spike-incomplete-section-pattern

### DIFF
--- a/app/controllers/users/otp_controller.rb
+++ b/app/controllers/users/otp_controller.rb
@@ -65,7 +65,7 @@ class Users::OtpController < DeviseController
   end
 
   def user_params
-    params.require(:user).permit(:uuid, :otp, :email)
+    params.require(:user).permit(:uuid, :otp, :email).tap { |p| p[:otp].strip! }
   end
 
   def new_referral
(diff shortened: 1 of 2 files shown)


vassyz (approved):
Does this actually happen? I mentioned it in a comment on Trello that there doesn't seem to be a space in the email. Another thing we could do would be to validate the code in `Users::OtpForm` to only have 6 digits.

paulodeon:
Honestly you must be a COD player with such mouse accuracy :)

But isn't it better UX just to strip whitespace than to display a validation error?

I do agree that a `/^\d{6}$/` regex validation would be better than the current one

vassyz:
I am a COD player, but on console 😄 

I think it's a good idea, but I can't replicate it. I double click on the code, but even when I use the mouse, I can't copy a space unless I copy `Get help` as well.

PR #616, opened 2023-03-24
Send the referral reminder email

vassyz:
### Context

This job will run every day at 9 am and send emails to the users who have a draft referral 7 days prior to deletion.

### Email preview

Screenshot 2023-03-27 at 12 11 02

### Link to Trello card

https://trello.com/c/ZPmG9xTb/1325-referral-reminder-email

### DIFF
--- a/app/jobs/draft_referrals_reminder_job.rb
+++ b/app/jobs/draft_referrals_reminder_job.rb
@@ -1,0 +1,19 @@
+class DraftReferralsReminderJob < ApplicationJob
+  sidekiq_options queue: "low"
+
+  def perform
+    return unless send_reminder_email?
+
+    Referral.stale_drafts_reminder.each do |referral|
+      next unless referral.reminder_emails.recent.none?
+
+      SendReminderEmailJob.perform_later(referral)
+    end
+  end
+
+  private
+
+  def send_reminder_email?
+    FeatureFlags::FeatureFlag.active?(:send_reminder_email)
+  end
+end
--- a/app/jobs/send_reminder_email_job.rb
+++ b/app/jobs/send_reminder_email_job.rb
@@ -1,0 +1,20 @@
+class SendReminderEmailJob < ApplicationJob
+  sidekiq_options queue: "low"
+
+  def perform(referral)
+    Referral.transaction do
+      record_reminder(referral)
+      send_reminder_email(referral)
+    end
+  end
+
+  private
+
+  def send_reminder_email(referral)
+    UserMailer.draft_referral_reminder(referral).deliver_later
+  end
+
+  def record_reminder(referral)
+    referral.reminder_emails.create!
+  end
+end
--- a/app/models/referral.rb
+++ b/app/models/referral.rb
@@ -10,11 +10,13 @@ class Referral < ApplicationRecord
   has_one_attached :pdf
 
   has_many :evidences, -> { order(:created_at) }, class_name: "ReferralEvidence", dependent: :destroy
+  has_many :reminder_emails
 
   scope :employer, -> { joins(:eligibility_check).where(eligibility_check: { reporting_as: :employer }) }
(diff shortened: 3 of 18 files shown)


vassyz:
[Referral reminder email](https://trello.com/c/ZPmG9xTb/1325-referral-reminder-email)

app/models/referral.rb
```diff
@@ -15,6 +15,8 @@ class Referral < ApplicationRecord
   scope :member_of_public, -> { joins(:eligibility_check).where(eligibility_check: { reporting_as: :public }) }
   scope :submitted, -> { where.not(submitted_at: nil) }
   scope :stale_drafts, -> { where(submitted_at: nil).where("created_at < ?", 90.days.ago) }
+  scope :stale_drafts_reminder,
```
gpeng, line 18: I don't think we should use the date here as we'll either miss some of send duplicates if we have a problem with the scheduled job for some reason. I think the logic is more reliable if we get all of the drafts that are more than 83 days old and then send the email if they haven't had one. That way it's no problem to rerun, worst that happens is a user may get their reminder late.

app/jobs/draft_referrals_reminder_job.rb
```diff
@@ -0,0 +1,9 @@
+class DraftReferralsReminderJob < ApplicationJob
+  sidekiq_options queue: "low"
+
+  def perform
+    return unless HostingEnvironment.production?
+
+    Referral.stale_drafts_reminder.each { |referral| UserMailer.draft_referral_reminder(referral).deliver_later }
```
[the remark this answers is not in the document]
vassyz: I will look into this, but I don't think the job can fail in such a way that the emails can get sent multiple times.

app/jobs/draft_referrals_reminder_job.rb
```diff
@@ -0,0 +1,9 @@
+class DraftReferralsReminderJob < ApplicationJob
+  sidekiq_options queue: "low"
+
+  def perform
+    return unless HostingEnvironment.production?
+
+    Referral.stale_drafts_reminder.each { |referral| SendReminderEmailJob.perform_later(referral.id) }
```
gpeng, line 7: I think you can pass the `referral` into the job and avoid having to fish it out of the DB in the job.

app/views/user_mailer/draft_referral_reminder.erb
```diff
@@ -0,0 +1,11 @@
+Your referral of serious misconduct by a teacher has not been updated in a while.
+
+It’ll be deleted in 7 days on <%= 7.days.from_now.to_fs(:day_month_and_year) %>.
+
+If you would like to keep your referral, you must make changes or complete it.
```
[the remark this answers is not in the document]
gpeng: Discussing this offline. We can address this in a subsequent PR.

gpeng (approved):
Looks good now 👍

app/jobs/send_reminder_email_job.rb
```diff
@@ -0,0 +1,28 @@
+class SendReminderEmailJob < ApplicationJob
+  sidekiq_options queue: "low"
+
+  def perform(referral_id)
+    @referral ||= Referral.find(referral_id)
+
+    return unless send_reminder?
+
+    send_reminder_email
```
gpeng, line 9: If the `record_reminder` fails we'll send the email again on the retry. Would this be better in a transaction like:
```
Referral.transaction do
  record_reminder
  send_reminder_eamil
end
```
and then deliver_now in the `send_reminder_email`? If the email failed the transaction wouldn't commit and if the record failed the job would retry without having sent the email. I think 😵‍💫

app/jobs/draft_referrals_reminder_job.rb
vassyz: Just saw your message. I didn't do that because as far as I know Sidekiq recommends passing simple data types as arguments: https://github.com/sidekiq/sidekiq/wiki/Best-Practices

PR #625, opened 2023-03-27
Hide show user referral page before submission

vassyz:
### Context

The `users/referrals/:id` page should not be accessible before the referral gets submitted.

Before:
Screenshot 2023-03-27 at 10 15 58

After:
Screenshot 2023-03-27 at 15 34 57

### Link to Trello card

https://trello.com/c/L7kl0DFc/1340-hide-show-user-referral-page-for-unsubmitted-referrals

### DIFF
--- a/app/controllers/users/referrals_controller.rb
+++ b/app/controllers/users/referrals_controller.rb
@@ -9,6 +9,7 @@ module Users
     end
 
     def show
+      redirect_to :not_found unless current_referral.submitted?
     end
 
     def current_referral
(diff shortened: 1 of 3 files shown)


app/controllers/users/referrals_controller.rb
```diff
@@ -9,6 +9,7 @@ def index
     end
 
     def show
+      redirect_to :not_found unless current_referral.submitted?
```
tavy88, line 12: Spec?

tavy88: But we should still have a spec I think...

tavy88 (approved):

PR #636, opened 2023-03-31
Refactor PreviousMisconductComponent

vassyz:
https://trello.com/c/CIr7zEGo/1337-epic-view-component-refactor

### DIFF
--- a/app/components/previous_misconduct_component.rb
+++ b/app/components/previous_misconduct_component.rb
@@ -1,63 +1,38 @@
 class PreviousMisconductComponent < ViewComponent::Base
   include ActiveModel::Model
   include ApplicationHelper
-  include ComponentHelper
   include ReferralHelper
 
   attr_accessor :referral
 
   def rows
-    items = [
-      {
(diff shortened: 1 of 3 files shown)


app/components/previous_misconduct_component.rb
```diff
@@ -1,63 +1,38 @@
...
-      }
-    end
+    summary_rows [previous_misconduct_reported_row, detailed_account_type_row, detailed_account_report_row].compact
+  end
+
+  def previous_misconduct_reported_row
+    {
+      label: "Has there been any previous misconduct?",
+      visually_hidden_text: "if there has been any previous misconduct",
+      value: nullable_value_to_s(humanize_three_way_choice(referral.previous_misconduct_reported)),
```
tavy88, line 16: No need to use `nullable_value_to_s` here. It's done by @paulodeon 's helper already. So it should be:

```ruby
value: humanize_three_way_choice(referral.previous_misconduct_reported)
```
here.

PR #648, opened 2023-04-05
Refactor the WhatHappened component

tavy88:
### Context

Refactor the WhatHappened component

Refactor the WhatHappened component

Screenshot 2023-04-05 at 11 07 33

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


vassyz (approved):

app/components/what_happened_component.rb
```diff
@@ -6,58 +6,37 @@ class WhatHappenedComponent < ViewComponent::Base
...
+    }
   end
 
-  def return_to
-    polymorphic_path([:edit, referral.routing_scope, referral, :allegation, :check_answers])
+  def dbs_notified_row
+    {
+      label: "Have you told DBS?",
+      visually_hidden_text: "if you have told DBS",
+      value: nullable_boolean_to_s(allegation_details(referral)),
```
tavy88, line 38: @vassyz we already discussed on this -> I think we need a separate ticket to investigate and refactor/remove `nullable_boolean_to_s`

app/helpers/referral_helper.rb
```diff
@@ -121,7 +121,7 @@ def summary_row(path:, label:, value:, visually_hidden_text: nil)
         text: label
       },
       value: {
-        text: nullable_value_to_s(value.presence)
```
tavy88, line 124: This is not exactly correct here because `false.presence == nil` -> causing the system to respond with `Not answered` instead of `No`.

PR #652, opened 2023-04-06
Refactor TheirRoleComponent

vassyz:
https://trello.com/c/CIr7zEGo/1337-epic-view-component-refactor

### DIFF
--- a/spec/components/their_role_component_spec.rb
+++ b/spec/components/their_role_component_spec.rb
@@ -1,0 +1,256 @@
+require "rails_helper"
+
+RSpec.describe TheirRoleComponent, type: :component do
+  subject(:component) { described_class.new(referral:) }
+
+  let(:referrer) { referral.referrer }
+  let(:row_labels) { component.rows.map { |row| row.dig(:key, :text) } }
+  let(:row_values) { component.rows.map { |row| row.dig(:value, :text) }.map(&:strip) }
+  let(:row_links) { component.rows.map { |row| row.dig(:actions, :href) } }
+
+  before { render_inline(component) }
+
+  context "without their role answers" do
+    let(:referral) { create(:referral, :complete) }
+
+    it "renders the labels" do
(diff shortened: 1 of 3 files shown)


spec/components/their_role_component_spec.rb
```diff
@@ -0,0 +1,256 @@
...
+
+    it "renders the values" do
+      expect(row_values).to eq(
+        [
+          "Not answered",
+          "Not answered",
+          "Not answered",
+          "Not answered",
+          "Not answered",
+          "Not answered"
```
tavy88, line 37: Is there much value in this test?

spec/components/their_role_component_spec.rb
```diff
@@ -0,0 +1,256 @@
...
+  end
+
+  context "when referral is submitted" do
+    let(:referral) { create(:referral, :submitted) }
+
+    it "does not have any action links" do
+      expect(row_links.compact).to be_empty
+    end
+  end
+end
```
tavy88, line 256: Because there are many potential combinations have you thought about having separate contexts for each conditional question in part? This way you could test all the logic per each question in part. I'll let you decide.

spec/components/their_role_component_spec.rb
vassyz: It tests that there is nothing being set by default and "Not answered" is rendered. But the reason I do that is that I can make sure the referral is empty from the factory, and then the values change when I explicitly set the attributes. I can remove it though, not that adamant to keep it.

tavy88: I don't think a user will ever get to this anyway because they can't really skip questions. Up to you.

vassyz: They can't skip the questions, but they can see the "Not answered" values if they don't answer the questions in the section, but click `Review and send` button.

PR #667, opened 2023-04-13
[1354] Download eligibility checks CSV

gpeng:
### Context

We need to monitor the effect of some changes to the eligibility checker but don't have the data in an easily accessible place currently.

* Add 'Download CSV' link to the eligibility checks index page

Stream the CSV as this is being generated from the transactional DB. As this data grows this page will start to timeout if we generate the whole CSV before responding.

Download link:

Screenshot 2023-04-13 at 17 02 29

Example CSV format:

Screenshot 2023-04-13 at 17 03 17

### DIFF
--- a/app/controllers/support_interface/eligibility_checks_controller.rb
+++ b/app/controllers/support_interface/eligibility_checks_controller.rb
@@ -1,7 +1,73 @@
+require "csv"
 module SupportInterface
   class EligibilityChecksController < SupportInterfaceController
     def index
-      @eligibility_checks = EligibilityCheck.order(updated_at: :desc).limit(100)
+      respond_to do |format|
+        format.html { @eligibility_checks = EligibilityCheck.order(updated_at: :desc).limit(100) }
+        format.csv { stream_csv }
+      end
+    end
+
+    private
+
+    def stream_csv
+      write_response_headers
+      write_csv_header_row
+      write_csv_rows
+    end
+
+    def write_response_headers
+      response.headers.delete("Content-Length")
+      response.headers["Cache-Control"] = "no-cache"
+      response.headers["Content-Type"] = "text/csv"
+      response.headers[
+        "Content-Disposition"
+      ] = "attachment; filename=eligibility-checks-#{Time.zone.now.to_i}.csv"
+    end
+
+    def write_csv_header_row
+      response.stream.write(
+        CSV.generate_line(
+          %w[
+            id
+            reporting_as
+            complained
+            is_teacher
+            teaching_in_england
+            serious_misconduct
+            unsupervised_teaching
+            eligibility_check_started
+            eligibility_check_last_updated
+            draft_referral_created
+            referral_submitted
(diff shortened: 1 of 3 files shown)


app/controllers/support_interface/eligibility_checks_controller.rb
```diff
@@ -1,7 +1,70 @@
...
+        )
+      end
+    end
+
+    def csv_row(check)
+      CSV.generate_line(
+        [
+          check.id,
+          check.reporting_as,
+          check.complained,
```
vassyz, line 58: `complained` is the only boolean value, and going to appear in the export as TRUE/FALSE instead of yes/no.

app/controllers/support_interface/eligibility_checks_controller.rb
```diff
@@ -1,7 +1,70 @@
...
+              eligbility_check_started
+              draft_referral_created
+              referral_submitted
+            ]
+        )
+      )
+    end
+
+    def write_csv_rows
+      EligibilityCheck.includes(:referral).order(:created_at).find_each do |check|
```
vassyz, line 46: Shouldn't we keep the same ordering on the website and the export. Either `updated_at` or `created_at`.

app/controllers/support_interface/eligibility_checks_controller.rb
```diff
@@ -1,7 +1,70 @@
...
+        CSV.generate_line(
+          %w[
+              Id
+              reporting_as
+              complained
+              is_teacher
+              teaching_in_england
+              serious_misconduct
+              unsupervised_teaching
+              eligbility_check_started
```
vassyz, line 37: `eligbility` typo

gpeng:
[Add CSV download link for eligibility checks.](https://trello.com/c/uqH6AMV7/1354-add-csv-download-link-for-eligibility-checks)

app/controllers/support_interface/eligibility_checks_controller.rb
```diff
@@ -1,7 +1,70 @@
+require "csv"
 module SupportInterface
   class EligibilityChecksController < SupportInterfaceController
     def index
-      @eligibility_checks = EligibilityCheck.order(updated_at: :desc).limit(100)
+      respond_to do |format|
+        format.html { @eligibility_checks = EligibilityCheck.order(updated_at: :desc).limit(100) }
+        format.csv { stream_csv }
```
tavy88, line 7: It would be good to have maybe a test scenario for this, unless it's urgent. Thoughts?

gpeng: Yeah. I've added a basic system spec now. It's not the best but the file download is a bit of a pain to test and I'm not sure the value we'd get from that is worth it. I'm having a go at a request spec and will add that in a separate PR if I get it working in 'not too long ™️'

tavy88 (approved):

PR #681, opened 2023-04-18
Revert "Bump dfe-analytics from v1.7.0 to v1.8.0"

vassyz:
Reverts DFE-Digital/refer-serious-misconduct#679


thomasleese (approved):
We had the same issue on Apply for QTS: https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/1300

PR #682, opened 2023-04-18
Extract sections and groups into their own class

paulodeon:
Promote form sections and groups into first class objects, and implement the About you and Organisation sections.

These sections still sit alongside the old Struct based sections with the idea that we will replace those with later PRs.

Once this is complete we can implement the Incomplete section pattern. 

Based on a simplification of https://github.com/DFE-Digital/refer-serious-misconduct/pull/611

One issue we found was that by using `valid?` on the form objects we were triggering the `after_validation` validation tracker so we had to apply a workaround for that.


app/forms/referrals/section.rb
```diff
@@ -0,0 +1,73 @@
...
+      items.find(&:incomplete?).path
+    end
+
+    def status
+      complete? ? :completed : :incomplete
+    end
+
+    def complete?
+      !!complete
+    end
```
tavy88, line 51: Can we make this a bit more readable?

app/forms/referrals/section.rb
```diff
@@ -0,0 +1,73 @@
...
+
+    def incomplete?
+      !complete
+    end
+
+    def questions_complete?
+      items[0..-2].all?(&:complete?)
```
tavy88, line 58: It would be great to add a comment on why this is needed.

app/forms/referral_form_section.rb
```diff
@@ -7,5 +7,32 @@ module ReferralFormSection
...
+    def slug
+      raise NotImplementedError, "You must define slug in #{self.class}"
+    end
+
+    def complete?
+      self.validation_tracking = false
+      result = valid?
+      self.validation_tracking = true
+      result
+    end
```
tavy88, line 24: Don't forget to add a comment here as well - we already discussed it.

app/forms/referrals/section.rb
paulodeon: I think this will be obvious once we drop in the incomplete section code, so will leave for now

tavy88 (commented):
Looks much cleaner! It LGTM - the last thing I would mention is that we have a few new classes that are not unit tested. Although their logic might be tested as part of other tests, I personally prefer having them unit tested. Otherwise, our test coverage will drop too. But I will let you decide how much testing is too much. Thoughts @gpeng @vassyz ?

app/forms/referral_form.rb
```diff
@@ -22,50 +22,31 @@ def submit
...
-          )
-        end
-      end
+    Referrals::SectionGroup.new(
+      slug: "about_you",
+      items: [
+        Referrals::Sections::ReferrerSection.new(referral:),
+        referral.from_employer? && Referrals::Sections::OrganisationSection.new(referral:)
+      ].compact_blank
+    )
```
tavy88, line 49: This could be simplified like this: 
```ruby
items = [Referrals::Sections::ReferrerSection.new(referral:)]
items << Referrals::Sections::OrganisationSection.new(referral:) if referral.from_employer?

Referrals::SectionGroup.new(
      slug: "about_you",
      items: items
    )
```

What do you think?

app/controllers/referrals/referrer/check_answers_controller.rb
```diff
@@ -4,7 +4,11 @@ class CheckAnswersController < BaseController
       def edit
         @referrer = current_referral.referrer
         @check_answers_form =
```
vassyz, line 6: Is it worth moving this to a method as it seems identical to the one in `update`?

vassyz:
I think we should have a separate branch for this feature and merge into it. I don't feel comfortable merging this into `main`. There are a few unused methods that shouldn't be added if they are not called yet. 

e.g. `questions_complete?` - not called anywhere and it runs `items[0..-2]` but in `OrganisationSection` `items` has only one element. I'm not sure how this will work, maybe it's correct, but the code is not there yet.

or

```ruby
view_component(user:)
        OrganisationComponent.new(referral:, user:, section: self)
```

It doesn't seem to be used and the `user` and `section` attributes are not set in that component.

These are also linear sections. I thing we shouldn't merge until we can see it working properly on a more complex logic, like Their Role.

app/controllers/referrals/referrer/check_answers_controller.rb
beano-digital: I tried that, but it actually isnt identical! the update is taking the param

paulodeon: I just tried this but they aren't actually identical! the update method takes a param

paulodeon:
So after @vassyz comments I realised we were in fact changing the functionality - this is no longer the case - and I've added a test to confirm that behaviour.

Also I added the check answers form to the organisation section which was missing.

Finally I removed the unneeded view_component and section_class stuff. See my changes in the Changes commit - I'll flatten it once we get approval

app/controllers/referrals/referrer/check_answers_controller.rb
vassyz: Missed that, then maybe we can sort the arguments to make it more obvious. Something like:

```ruby
referral: current_referral,
referrer: current_referral.referrer,
complete: @referrer.complete
```
and
```ruby
referral: current_referral,
referrer: current_referral.referrer,
complete: referrer_params[:complete]
```

PR #683, opened 2023-04-19
Add missing specs for pagination

tavy88:
### Context

Add missing specs for pagination

### DIFF
--- a/spec/factories/staffs.rb
+++ b/spec/factories/staffs.rb
@@ -21,4 +21,9 @@ FactoryBot.define do
   trait :can_manage_referrals do
     manage_referrals { true }
   end
+
+  trait :random do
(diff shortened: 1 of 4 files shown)


spec/factories/staffs.rb
```diff
@@ -21,4 +21,9 @@
   trait :can_manage_referrals do
     manage_referrals { true }
   end
+
+  trait :random do
+    email { Faker::Internet.email }
```
vassyz, line 26: We are using this pattern for users `email { "#{SecureRandom.hex(5)}@example.com" }` I don't think it's worth adding the `faker` gem just for that.

tavy88: Ok I can do that instead.

vassyz (approved):

PR #695, opened 2023-04-24
Hide "Not known" rows when the previous answer is "No"

vassyz:
They were showing up as "Not known" or "Not answered" when the previous answer was "No". This change makes the referral forms consistent with the manage interface.

Before:
Screenshot 2023-04-24 at 18 03 33

After:
Screenshot 2023-04-24 at 18 03 16

I also updated the "Their name" row because it needs to be `nil` to show a "Not answered" row. If the name was missing the interpolation would return an empty string.

Before:
Screenshot 2023-04-24 at 18 05 38

After:
Screenshot 2023-04-24 at 18 04 38

### DIFF
--- a/app/components/contact_details_component.rb
+++ b/app/components/contact_details_component.rb
@@ -10,12 +10,12 @@ class ContactDetailsComponent < ViewComponent::Base
     items =
       summary_rows [
                      email_known_row,
-                     email_row,
+                     referral.email_known? && email_row,
                      phone_number_known_row,
-                     phone_number_row,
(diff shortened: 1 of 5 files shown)


vassyz:
[Hide "Not known" rows when the previous question the referrer answered "No"](https://trello.com/c/E9ejWQ7I/1358-hide-not-known-rows-when-the-previous-question-the-referrer-answered-no)

paulodeon (approved):
LGTM, although there probably other "known" row cases. If you're able to include those as well it would be helpful to my PR

PR #701, opened 2023-04-26
Remove log file size limiter and fix typos

vassyz:
I noticed that the `test.log` on my local was over 800Mb. I looked into it as I knew I added a 50Mb limit. It seems that adding `rails_semantic_logger` will not use the environment setting anymore. I couldn't find a solution nor a setting in the official documentation https://logger.rocketjob.io/rails. In the specs the gem is only used for testing the `LogStashFormatter`. This is a mild annoyance as every test suite run adds another 10-15Mb to the log file.

I removed it for development as well as the log filesize is tiny in comparison, and Rails 7.1 should have a new setting `log_file_size ` set to 100Mb by default. https://github.com/rails/rails/pull/44888

I also fixed some random typos.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -74,7 +74,4 @@ Rails.application.configure do
           end
         end
     )
-
-  # Limit logs to 50MB
-  config.logger = ActiveSupport::Logger.new(config.paths["log"].first, 1, 50.megabytes)
 end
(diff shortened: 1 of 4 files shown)


paulodeon:
Another idea here would be to disable semantic logger in dev and test - personally I find it hard to read compared to the good ol rails logger. Wdy think?

paulodeon (approved):

PR #698, opened 2023-04-25
Incomplete section pattern

paulodeon:
https://trello.com/c/UzOXyxpo/1346-update-the-ui-to-display-completeness-and-link-to-the-correct-questions

PR #710, opened 2023-04-28
Send referral to all check answers controllers

paulodeon:
Extraction from #698


vassyz (approved):
Linter is failing though

PR #713, opened 2023-05-02
Refactor form attributes

vassyz:
Adds a `CustomAttrs` module that changes setting instance variables for forms from defining methods to using attributes.


app/forms/referral_form_section.rb
```diff
@@ -50,5 +50,32 @@ def section_class
...
+      attr_object(:referrer, *args)
+    end
+
+    def self.attr_object(obj, *args)
+      args.each do |attr_name|
+        inst_variable_name = "@#{attr_name}".to_sym
+        define_method(attr_name) do
+          var = instance_variable_get(inst_variable_name)
+
+          return var unless var.nil?
```
paulodeon, line 72: I think this should be `return var if instance_variable_defined?(inst_variable_name)`. This allows us to handle nils

app/forms/referral_form_section.rb
```diff
@@ -50,5 +50,32 @@ def section_class
...
+
+          attr_value = try(obj).try(attr_name)
+          instance_variable_set(inst_variable_name, attr_value)
+          attr_value
+        end
+      end
+    end
```
paulodeon, line 79: Why not define the writer here as well? Would allow us to get rid of all those below

paulodeon: If we could also add a `:boolean` option here and add `ActiveModel::Type::Boolean.new.cast(value)` that would really clean things up

paulodeon (approved):

app/forms/referral_form_section.rb
vassyz: How is this any different than `var.nil?`?

paulodeon: See https://www.justinweiss.com/articles/4-simple-memoization-patterns-in-ruby-and-one-gem/ for an explanation

app/forms/referral_form_section.rb
vassyz: Wanted to do this in a separate PR, because this wasn't part of the refactor. Those type casting methods were already there. I'll take a look though.

app/forms/referral_form_section.rb
vassyz: I get that, but your original PR I'm basing this on is not doing this. I'm just replicating the same behaviour. Let's have a chat.

vassyz:
Made redundant by #717

PR #714, opened 2023-05-02
Group eligibility screener controllers into a folder

vassyz:
Makes is easier to add new pages in the future.


config/routes.rb
```diff
@@ -45,21 +45,21 @@
     get "/start", to: "pages#start"
     get "/users/registrations/exists", to: "users/existing_registration#new"
     post "/users/registrations/exists", to: "users/existing_registration#create"
-    get "/referral-type", to: "referral_type#new"
-    post "/referral-type", to: "referral_type#create"
-    get "/have-you-complained", to: "have_complained#new"
-    post "/have-you-complained", to: "have_complained#create"
+    get "/referral-type", to: "eligibility_screener/referral_type#new"
```
paulodeon, line 48: Anything against using resource routes here i.e. 
```ruby
    namespace :eligibility_screener, path: 'eligibility-screener' do
      resource :referral_type, path: 'referral-type', only: %i[new create]
    end
```

paulodeon (approved):
Good spot - I had my eye on this refactor, definitely cleans it up a bit and makes it easier to understand

config/routes.rb
vassyz: I didn't want to change the URLs

paulodeon: But could be done in another PR

vassyz: If you keep the resource, the convention is to have `/new` for new, so I'd still have to change the URLs.

paulodeon: I see

PR #731, opened 2023-05-11
Fix back links on the eligibility pages

vassyz:
The back links were using the `url_for(:back)` method which was causing the back link to go back to the previous page in the browser history rather than the previous page in the eligibility flow.


vassyz:
[Fix back links on the eligibility pages](https://trello.com/c/zzhEZo9P/1370-fix-back-links-on-the-eligibility-pages)

paulodeon (approved):
LGTM

PR #732, opened 2023-05-11
Use CustomAttrs for eligibility forms

vassyz:
Updates the eligibility pages to use custom attributes. This will also help with showing the selected answers when navigating back to the page and automatically converting true/false to boolean values.

It also fixes a bug where calling `process_type` for a db string value that get processed as a symbol. E.g. the `reporting_as` values of :public, :employer

### DIFF
--- a/app/forms/eligibility_screener/have_complained_form.rb
+++ b/app/forms/eligibility_screener/have_complained_form.rb
@@ -1,16 +1,8 @@
 module EligibilityScreener
-  class HaveComplainedForm
-    include ActiveModel::Model
+  class HaveComplainedForm < EligibilityScreenerForm
+    attr_eligibility_check :complained
 
-    attr_accessor :eligibility_check
-    attr_reader :complained
-
-    validates :eligibility_check, presence: true
     validates :complained, inclusion: { in: [true, false] }
-
-    def complained=(value)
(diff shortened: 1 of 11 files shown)


app/forms/eligibility_screener/have_complained_form.rb
```diff
@@ -1,17 +1,14 @@
 module EligibilityScreener
   class HaveComplainedForm
     include ActiveModel::Model
+    include CustomAttrs
```
paulodeon, line 4: While you're here is it worth extracting some of this into a base class or a module for all eligibility forms?

vassyz: You mean a class that just does `include CustomAttrs`? Is it worth doing that?

paulodeon: it looks like it would also include

```   
include ActiveModel::Model
include CustomAttrs

attr_accessor :eligibility_check
validates :eligibility_check, presence: true
```

paulodeon (approved):

app/forms/eligibility_screener/have_complained_form.rb
vassyz: I think it's worth doing.

PR #734, opened 2023-05-12
Use custom attributes on check answers forms

vassyz:


app/forms/referrals/allegation/check_answers_form.rb
```diff
@@ -3,15 +3,12 @@ module Referrals
   module Allegation
     class CheckAnswersForm
       include ReferralFormSection
+      include CustomAttrs
```
paulodeon, line 6: Can we put this inside `ReferralFormSection`

paulodeon (approved):

app/forms/referrals/allegation/check_answers_form.rb
paulodeon: Then at some point we can rename `ReferralFormSection` and make it a base class as well

vassyz: It's already in the `ReferralFormSection`. I realised that when I saw your comment on the other PR, but forgot to remove. Updated now.

PR #758, opened 2023-05-26
Refactor file upload to use Upload model

vassyz:
This commit refactors the file upload functionality to use the Upload model. This is a polymorphic model that can be attached to any other model. It is used to store the file data and metadata.

I also added a rake task to transfer the existing file data to the new model.

### DIFF
--- a/app/models/referral.rb
+++ b/app/models/referral.rb
@@ -1,10 +1,12 @@
 class Referral < ApplicationRecord
+  include Uploadable
+
   belongs_to :eligibility_check, dependent: :destroy
   belongs_to :user
 
   has_one :organisation, dependent: :destroy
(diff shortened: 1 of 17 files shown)


app/models/referral.rb
```diff
@@ -1,10 +1,12 @@
 class Referral < ApplicationRecord
+  include Uploadable
```
paulodeon, line 2: This could be made more DSL like, so you could do:

```ruby
include Uploadable

has_one_uploadable :allegation
```

and then get all the relevant methods etc?

gpeng (commented):
Something along these lines looks good to me

vassyz:
Superseded by #764

PR #766, opened 2023-06-02
Add /start-govuk route

gpeng:
GOV.UK doesn't pass the referrer so we can't measure how many users are coming via the correct start page vs the other links on there. Until we can rationalise the links we can set it up to link to a specific URL for analysis.

Add `/start-govuk` route. This just redirects to `pages#start` which funnels the user to the correct first page based on whether they're logged in or not. We'll only use this for the GOV.UK start page.


vassyz (approved):

gpeng:
Closing as we've been unable to update the GOV.UK link to use this path thus far.

PR #767, opened 2023-06-05
Allow sign in to manage with AD

paulodeon:
See README.md for more info


paulodeon:
A chat with @gpeng revealed that they do a similar thing differently at publish: https://github.com/DFE-Digital/publish-teacher-training/blob/85f2ba87a7277c38e54590b9d8f611e0c8755347/app/services/authentication_service.rb#L51 

They don't use devise though so have to do some of the work themselves.

They also handle email changes which we currently do not and would need to deal with before merging

PR #781, opened 2023-06-12
Add ability to view validation errors in the manage panel

paulodeon:
Based on feedback we need more detail in the error history - this provides that


vassyz (approved):
Looks good, no specs though.

PR #823, opened 2023-07-13
Disambiguate filenames in zip file

gpeng:
We create a zip file of uploads from a referral to allow transfer into the case management system. The existing implementation has a bug where if multiple files have the same name in a section a `Zip::EntryExistsError
` is raised.

* Add an incrementing index to the file names as we're creating the downloadable zip file.

This handles the case where the user has uploaded identically named files to the same section. We could probably handle this more gracefully as this implementation adds an index to every file name when it is very rarely required. That could be a further iteration but this fixes a production issue.


AbigailMcP (approved):
🚀

PR #922, opened 2023-11-27
seed staff user in review apps

asmega:


richardpattinson (approved):
Looks great, useful stuff

asmega:
> Looks great, useful stuff

Doesn't actually work. not sure why. also since i don't have access i can't pinpoint why

asmega:
not needed for now

PR #954, opened 2023-12-14
Review wording for adding feedback link from refer page to email content

richardpattinson:
Edit the content on the submit feedback link. Add the link to the email confirming that a referral has been submitted.


asmega (commented):
You've got failing tests. So in mailers you don't want to be using `*_path`, you'll want to use `*_url` because in the email it needs the full url including the domain. otherwise you'll  end up with `gmail.com/the/path/you/wanted`

see https://guides.rubyonrails.org/action_mailer_basics.html#generating-urls-in-action-mailer-views for more detail

PR #955, opened 2023-12-14
Build of evidence upload changes

richardpattinson:
Currently the only specified criteria for evidence upload is that it must be smaller than 50MB.  Given validation errors around this area are high and people are unaware, it would be good to look at the language on this page and amend to make it clearer what evidence is accepted to prevent errors.


asmega (approved):
Looks fine but you need to rebase. that first commit is already applied to `main`

PR #964, opened 2024-01-02
enable sentry performance

asmega:
# Context

- We want to be able to track performance so we can identify any bottlenecks or any issues that creep in

# Changes

- Enabled tracing for sentry to track performance
- It samples 10% percent of requests and 0.1% of the health check requests

### DIFF
--- a/config/initializers/sentry.rb
+++ b/config/initializers/sentry.rb
@@ -20,4 +20,12 @@ Sentry.init do |config|
     "ActionDispatch::Http::Parameters::ParseError",
     "Mime::Type::InvalidMimeType"
   ]
+
+  config.traces_sampler = lambda do |sampling_context|
+    rack_env = sampling_context[:env]
+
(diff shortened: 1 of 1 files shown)


richardpattinson (commented):
I dont really understand what this is doing; how does this enable sentry permornance monitoring?

asmega:
> I dont really understand what this is doing; how does this enable sentry permornance monitoring?

Yep so in sentry under performance we now see some metrics for this review app in there, so this does enable it. 

The docs on how to configure can be found here https://docs.sentry.io/platforms/ruby/guides/rails/performance/?original_referrer=https%3A%2F%2Fwww.google.com%2F#configure So it looks like by default it is not enabled which is why prior to this change you don't see any data. it accepts boolean or some kind of decimal value. So for the health check where we don't really care about performance I just put that to 0.1%, the remainder of the requests, 10% will get profiled.

richardpattinson (approved):

PR #973, opened 2024-01-08
Update email about deleting drafts in 90 days

asmega:
# Context

- We delete draft referrals but don't inform users of this in corresponding emails

# Changes

- Updated email copy when user creates a draft referral, it now includes notice of 90 days of inactivity leads to deletion


richardpattinson (approved):
Looks good to me

PR #985, opened 2024-01-16
Implement new MOP journey

richardpattinson:
Context

This is the build step to update the MOP eligibility checker journey

Resources

Prototype is here 

Ticket RE titles, url and validation 

Changes

Start page changes are out of scope for this ticket as under GDS control

See checklists below…

Technical considerations

There is a lot of inheritance between MOP and employer journeys. This was probably done this way as the journeys are the same.

Given we are diverging these you may want to cull the inheritance first and opt for a greater degree of duplication leading to more flexibility in return. I recommend this step is done first otherwise you’ll be going backwards and forwards and probably eat up more time

ACs

New updated MOP eligibility checker journey matches prototype.

Titles, URLs, validation match up with the outcome of


app/views/public_eligibility_screener/is_teacher/new.html.erb
```diff
@@ -0,0 +1,29 @@
...
+<% content_for :back_link_url, previous_question_path %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with model: @is_teacher_form, url: public_is_a_teacher_url, method: :post do |f| %>
+      <%= f.govuk_error_summary %>
+
+      <h1 class="govuk-heading-l">Who the allegation is about</h1>
+
+      <p>You can only refer a teacher if they’ve been engaged in unsupervised teaching work, such as:</p>
```
asmega, line 11: `<p>` tags are missing `class="govuk_body"`

app/views/public_eligibility_screener/allegation/new.html.erb
```diff
@@ -0,0 +1,23 @@
+<% content_for :page_title, "Does your allegation involve any of the following?" %>
```
asmega, line 1: Compared to the other forms missing the `Error: ` prefix

app/forms/public_eligibility_screener/teaching_in_england_form.rb
```diff
@@ -0,0 +1,14 @@
+module PublicEligibilityScreener
+  class TeachingInEnglandForm < PublicEligibilityScreenerForm
+    attr_eligibility_check :teaching_in_england
+
+    validates :teaching_in_england, inclusion: { in: %w[yes no not_sure] }
+
+    def save
+      return false unless valid?
+
+      eligibility_check.teaching_in_england = teaching_in_england
```
asmega, line 10: might get the issue we saw together, where `@teaching_in_england` should be used over `teaching_in_england`

app/forms/public_eligibility_screener/reporting_as_form.rb
```diff
@@ -0,0 +1,14 @@
+module PublicEligibilityScreener
+  class ReportingAsForm < PublicEligibilityScreenerForm
+    attr_eligibility_check :reporting_as
+
+    validates :reporting_as, presence: true
+
+    def save
+      return false unless valid?
+
+      eligibility_check.reporting_as = reporting_as
```
asmega, line 10: might get the issue we saw together, where `@reporting_as` should be used over `reporting_as`

app/forms/public_eligibility_screener/is_teacher_form.rb
```diff
@@ -0,0 +1,14 @@
+module PublicEligibilityScreener
+  class IsTeacherForm < PublicEligibilityScreenerForm
+    attr_eligibility_check :is_teacher
+
+    validates :is_teacher, inclusion: { in: %w[yes no not_sure] }
+
+    def save
+      return false unless valid?
+
+      eligibility_check.is_teacher = is_teacher
```
asmega, line 10: might get the issue we saw together, where `@is_teacher` should be used over `is_teacher`

app/controllers/concerns/public_enforce_question_order.rb
```diff
@@ -0,0 +1,118 @@
...
+
+  def serious_misconduct_answered?
+    !eligibility_check.serious_misconduct.nil?
+  end
+
+  def continue_with_answered?
+    !eligibility_check.continue_with.nil?
+  end
+
+  def handle_multiple_branches
```
asmega, line 109: minor: I'd probably rename this to `complained_to_school_question`

app/controllers/public_eligibility_screener/allegation_controller.rb
```diff
@@ -0,0 +1,25 @@
+module PublicEligibilityScreener
+  class AllegationController < PublicEligibilityScreenerController
+    def new
+      @allegation_form = PublicEligibilityScreener::AllegationForm.new(eligibility_check:)
+    end
+
+    def create
+      @allegation_form =
+        PublicEligibilityScreener::AllegationForm.new(allegation_params.merge(eligibility_check:))
+      if allegation_params[:serious_misconduct] == ["","none"]
```
asmega, line 10: so we want to move this from the controller to the form. so we populate the form with the state (params) and house the logic there

app/controllers/public_eligibility_screener/allegation_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+        redirect_to public_consider_if_you_should_make_a_referral_path
```
asmega, line 11: I'm guessing this branching will go once we trigger the correct errors on the form and it will look simpler with something along the lines of
```ruby
if @form.save
  redirect_to_next_question
else
  render :new
end
```

app/views/public_eligibility_screener/unsupervised_teaching/new.html.erb
```diff
@@ -0,0 +1,31 @@
...
+
+      <ul class="govuk-list govuk-list--bullet">
+        <li>planning and preparing lessons and courses for pupils</li>
+        <li>delivering lessons to pupils</li>
+        <li>assessing the development, progress and attainment of pupils</li>
+        <li>reporting on the development, progress and attainment of pupils</li>
+      </ul>
+
+      <%= f.govuk_radio_buttons_fieldset :unsupervised_teaching, legend: { size: 'm', text: 'Does this person carry out unsupervised teaching work?' } do %>
+        <%= f.hidden_field :unsupervised_teaching %>
```
asmega, line 23: why do we always seem to add a hidden field?

richardpattinson: I am not sure what it does; but if it breaks if i remove it

app/controllers/concerns/public_enforce_question_order.rb
```diff
@@ -0,0 +1,118 @@
...
+  def referral_type_answered?
+    !eligibility_check.reporting_as.nil?
+  end
+
+  def unsupervised_teaching_answered?
+    !eligibility_check.unsupervised_teaching.nil?
+  end
+
+  def is_allegation_none_of_the_options?
+    eligibility_check.serious_misconduct == "[\"none\"]"
```
asmega, line 86: i know this partially works for now. but i think we want to structure this with a better data structure in the future

app/views/public_eligibility_screener/awaiting_outcome/new.html.erb
```diff
@@ -0,0 +1,20 @@
+<% content_for :page_title, "Wait for an outcome before making a referral" %>
+<% content_for :back_link_url, public_complained_to_school_path %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+    <h1 class="govuk-heading-l">Wait for an outcome before making a referral</h1>
+    <p class="govuk_body">
+      Check the school's complaint policy for the processing time. It's usually available on the school's website or from the school office.
```
asmega, line 8: fyi you'll want to convert all apostrophes to smart quotes which are more readable

asmega (approved):

PR #1002, opened 2024-01-30
Condense Validation Upload Size Errors

richardpattinson:
Add wrapper in refer-serious-misconduct/app/models/concerns/validation_tracking.rb that checks in the value for field method that checks for ant files over size limit. This will allow support staff to quickly assess how many files in a batch upload are over the file size limit.

### DIFF
--- a/app/lib/upload_wrapper.rb
+++ b/app/lib/upload_wrapper.rb
@@ -1,14 +1,16 @@
 class UploadWrapper
   include ActiveSupport::NumberHelper
 
-  attr_reader :upload
+  attr_reader :upload, :count
 
-  def initialize(upload:)
+  def initialize(upload:, count:)
     @upload = upload
+    @count = count
   end
 
   def as_json(*args)
     r = super
-    r.merge(file_size: number_to_human_size(upload.size, strip_insignificant_zeros: false))
+    r.merge(file_size: number_to_human_size(upload.size, strip_insignificant_zeros: false),
+            files_with_errors_count: count)
   end
 end
--- a/app/models/concerns/validation_tracking.rb
+++ b/app/models/concerns/validation_tracking.rb
@@ -44,9 +44,10 @@ module ValidationTracking
     value = public_send(field)
 
     if value.instance_of?(Array)
+      count = value.count {|file| file.size >= FileUploadValidator::MAX_FILE_SIZE }
       value = value.map do |upload|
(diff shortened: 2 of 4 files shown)


app/lib/upload_wrapper.rb
```diff
@@ -9,6 +9,14 @@ def initialize(upload:)
 
   def as_json(*args)
     r = super
-    r.merge(file_size: number_to_human_size(upload.size, strip_insignificant_zeros: false))
+    r.merge(file_size: number_to_human_size(upload.size, strip_insignificant_zeros: false, files_over_size: files_over_size))
+  end
+
+  def files_over_size
+    count = 0
+    @upload.each do |file|
```
asmega, line 17: this wrapper as it stands only ever wraps a single upload hence the singular name. so i'm pretty sure you won't be able to count the number of uploads here?

app/lib/upload_wrapper.rb
```diff
@@ -3,12 +3,13 @@ class UploadWrapper
...
-  def initialize(upload:)
+  def initialize(upload:, count:)
     @upload = upload
+    @count = count
   end
 
   def as_json(*args)
     r = super
-    r.merge(file_size: number_to_human_size(upload.size, strip_insignificant_zeros: false))
+    r.merge(file_size: number_to_human_size(upload.size, strip_insignificant_zeros: false), files_over_size_count: count)
```
asmega, line 13: - You'll end up with the `count` attached to every file uploaded, is this what you intended?
- `files_over_size_count` isn't technically the correct name for this as any error will on an upload will contribute towards this count which may lead it to be misleading. maybe `files_with_errors_count` would be a more accurate description?

asmega: do we want some tests to cover the new behaviour?

asmega:
probably want to fix failing tests and linting?

app/views/support_interface/validation_errors/index.html.erb
```diff
@@ -2,6 +2,12 @@
 
 <h1 class="govuk-heading-l">Validation errors</h1>
 
+<%= form_with model: @filter_form, url: support_interface_validation_error_filter_path(ValidationError.new()) , method: :get do |f| %>
```
asmega, line 5: i think you've accidentally pushed code from another branch here

asmega (approved):

PR #1019, opened 2024-02-13
Add filter to validation errors

richardpattinson:
Add filter to validation errors

### DIFF
--- a/app/controllers/support_interface/validation_errors_controller.rb
+++ b/app/controllers/support_interface/validation_errors_controller.rb
@@ -1,18 +1,25 @@
 module SupportInterface
   class ValidationErrorsController < SupportInterfaceController
     def index
-      @grouped_counts = ValidationError.group(:form_object).order("count_all DESC").count
-      @grouped_column_error_counts = ValidationError.list_of_distinct_errors_with_count
+      @filter_form = FilterValidationForm.new
     end
 
     def history
       @form_object = params[:form_object]
       @attribute = params[:attribute]
-
-      @errors = ValidationError.where(form_object: @form_object).order(created_at: :desc)
+      @errors =
+      if @form_object.present?
+        ValidationError.where(form_object: @form_object).order(created_at: :desc)
+      else
+        ValidationError.all
+      end
 
       @errors_with_attribute_extracted =
-        ValidationError.extract_attribute_from_errors(@errors, @attribute)
+        ValidationError.filter_on_attributes(@errors, @attribute)
+    end
+
+    def filter_params
+      params.require(:filter_validation_form).permit(:form_object, :attribute)
     end
   end
 end
--- a/app/models/validation_error.rb
+++ b/app/models/validation_error.rb
@@ -23,4 +23,23 @@ class ValidationError < ApplicationRecord
       end
     end
   end
+
+  def self.extract_all_attributes_from_errors(errors)
+    errors.filter_map do |error|
+      attribute = error.details.keys.first
+      OpenStruct.new(
+        created_at: error.created_at,
+        messages: error.details.dig(attribute, "messages"),
+        value: error.details.dig(attribute, "value")
+      )
+    end
+  end
+
+  def self.filter_on_attributes(errors, attribute)
+    if attribute.empty?
+      extract_all_attributes_from_errors(errors)
+    else
+      extract_attribute_from_errors(errors, attribute)
+    end
+  end
 end
--- a/app/views/support_interface/validation_errors/index.html.erb
+++ b/app/views/support_interface/validation_errors/index.html.erb
@@ -2,6 +2,12 @@
 
 <h1 class="govuk-heading-l">Validation errors</h1>
 
+<%= form_with model: @filter_form, url: history_support_interface_validation_errors_path, method: :get, scope: "" do |f| %>
+  <%= f.govuk_collection_select :form_object, @filter_form.form_objects_select, :id, :name, label: { text: "Filter by form" } %>
+  <%= f.govuk_collection_select :attribute, @filter_form.attribute_select, :id, :name, label: { text: "Filter by attribute" } %>
+  <%= f.govuk_submit %>
+<% end %>
+
 <table class="govuk-table" summary="Attribute error">
   <thead class="govuk-table__head">
     <tr class="govuk-table__row">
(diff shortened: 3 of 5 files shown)


app/controllers/support_interface/validation_errors_controller.rb
```diff
@@ -1,18 +1,37 @@
...
   class ValidationErrorsController < SupportInterfaceController
     def index
+      @filter_form = FilterValidationForm.new
       @grouped_counts = ValidationError.group(:form_object).order("count_all DESC").count
+
       @grouped_column_error_counts = ValidationError.list_of_distinct_errors_with_count
     end
 
+    def filter
+      redirect_to history_support_interface_validation_errors_path(form_object: params[:filter_validation_form][:form_object], attribute: filter_params[:filter_validation_form][:attribute])
```
asmega, line 11: you are redirecting back to the original resource, why the extra proxy and not directly `GET` `history` without the extra step

app/controllers/support_interface/validation_errors_controller.rb
```diff
@@ -1,18 +1,38 @@
...
+      @form_objects_select = @grouped_counts.map {|key, _value|
+ OpenStruct.new(id: key, name: humanized_form_object(key)) }.append(OpenStruct.new(id: "", name: "All"))
+
       @grouped_column_error_counts = ValidationError.list_of_distinct_errors_with_count
+      @attribute_select = @grouped_column_error_counts.map {|column|
+ OpenStruct.new(id: column[0][1], 
+name: column[0][1].underscore.titleize.gsub('.', ' ')) }.append(OpenStruct.new(id: "", name: "All"))
+    end
+
+    def filter
```
asmega, line 16: are we able to remove this controller action and plumb the form directly to `history`?

app/controllers/support_interface/validation_errors_controller.rb
```diff
@@ -1,18 +1,38 @@
...
+      redirect_to history_support_interface_validation_errors_path(form_object: filter_params[:form_object], 
+attribute: filter_params[:attribute])
     end
 
     def history
       @form_object = params[:form_object]
       @attribute = params[:attribute]
-
-      @errors = ValidationError.where(form_object: @form_object).order(created_at: :desc)
+      @errors = if !@form_object.empty?
```
asmega, line 24: are we able to not use a double negative?
```suggestion
      @errors = if @form_object.present?
```

app/controllers/support_interface/validation_errors_controller.rb
```diff
@@ -1,18 +1,38 @@
...
+        ValidationError.where(form_object: @form_object).order(created_at: :desc)
+      else
+        ValidationError.all
+                end
```
asmega, line 28: indentation of this block seems off?

app/views/support_interface/validation_errors/index.html.erb
```diff
@@ -2,6 +2,12 @@
 
 <h1 class="govuk-heading-l">Validation errors</h1>
 
+<%= form_with model: @filter_form, url: filter_support_interface_validation_errors_path, method: :get do |f| %>
+  <%= f.govuk_collection_select :form_object, @form_objects_select, :id, :name, label: { text: "Filter by form" } %>
```
asmega, line 6: given you've created a shiny new form object `FilterValidationForm` can this data be housed there rather than as instance variables in the controller and view? ie

```
class FilterValidationForm
  def grouped_counts
    @grouped_counts ||= ValidationError.group(:form_object).order("count_all DESC").count
  end

  def form_objects_select
    @form_objects_select ||= grouped_counts.map do |key, _value|
      OpenStruct.new(id: key, name: humanized_form_object(key))
    end.append(OpenStruct.new(id: "", name: "All"))
  end
end

<%= f.govuk_collection_select :form_object, @filter_form.form_objects_select, :id, :name, label: { text: "Filter by form" } %>
```

app/models/validation_error.rb
```diff
@@ -23,4 +31,35 @@ def self.extract_attribute_from_errors(errors, attribute)
...
+      attribute = error.details.keys.first
+      OpenStruct.new(
+        created_at: error.created_at,
+        messages: error.details.dig(attribute, "messages"),
+        value: error.details.dig(attribute, "value")
+      )
+    end
+  end
+
+  def self.filter_on_errors(form_object)
```
asmega, line 46: is this new method used somewhere? or am i missing it?

app/models/validation_error.rb
```diff
@@ -23,4 +31,35 @@ def self.extract_attribute_from_errors(errors, attribute)
...
+
+  def self.filter_on_attributes(errors, attribute)
+    if attribute.empty?
+      extract_all_attributes_from_errors(errors)
+    else
+      extract_attribute_from_errors(errors, attribute)
+    end
+  end
+
+  def self.list_form_objects
```
asmega, line 62: is this new method used somewhere? or am i missing it?

app/controllers/support_interface/validation_errors_controller.rb
```diff
@@ -1,18 +1,38 @@
 module SupportInterface
   class ValidationErrorsController < SupportInterfaceController
+    include ValidationErrorHelper
     def index
+      @filter_form = FilterValidationForm.new
       @grouped_counts = ValidationError.group(:form_object).order("count_all DESC").count
```
asmega, line 6: although not code you've introduced I'm pretty sure the `order` clause is redundant if we're performing a `count` after it as the ordering of the records will not affect the `count`

app/controllers/support_interface/validation_errors_controller.rb
richardpattinson: I think we can't move it directly to the history route; 
`<%= form_with model: @filter_form, url: filter_support_interface_validation_errors_path, method: :get do |f| %>
  <%= f.govuk_collection_select :form_object, @filter_form.form_objects_select, :id, :name, label: { text: "Filter by form" } %>
  <%= f.govuk_collection_select :attribute, @filter_form.attribute_select, :id, :name, label: { text: "Filter by attribute" } %>
  <%= f.govuk_submit %>
<% end %>`
is the block where we put in the route; we cant supply arguments to the path there as the params are yet to be selected.

richardpattinson: For example history_path(param1, param2) would fail if you input it as the url for that form with model block

asmega: this patch should work? https://github.com/DFE-Digital/refer-serious-misconduct/commit/5a209f764e2e98af781f117ff5acab1460ecb778

app/policies/support_policy.rb
```diff
@@ -38,4 +38,8 @@ def deactivate?
   def history?
     user.view_support?
   end
+
+  def filter?
+    user.view_support?
+  end
```
asmega, line 44: this is no longer needed as controller action has now been removed

asmega (approved):

PR #1038, opened 2024-02-23
Feedback backoffice

asmega:
# Context

- There is currently no back office system to be able to view feedback provided by users

# Changes

- Introduce ability to view feedback from users in the back office area
- Only available to `Staff` that have the role `view_support: true`

# Notes for review

- This is built atop https://github.com/DFE-Digital/refer-serious-misconduct/pull/1033 which should be reviewed and merged first. This is because it contains some changes which was needed for this PR


richardpattinson (approved):
Looks great, good to ship

PR #1046, opened 2024-03-04
Edit content to state zip files cant be uploaded

richardpattinson:
Users are getting error messages because they’re trying to upload zip files.

The hint in the upload file fields has been changed to advise that this cannot be done.

### DIFF
--- a/app/views/public_referrals/allegation_evidence/upload/edit.html.erb
+++ b/app/views/public_referrals/allegation_evidence/upload/edit.html.erb
@@ -34,7 +34,7 @@
           :evidence_uploads,
           multiple: true,
           label: { text: "Upload files", size: "m" },
-          hint: { text: "You can upload up to #{FileUploadValidator::MAX_FILES} files. Each file must be smaller than #{max_allowed_file_size}. Larger files may take longer to upload. We will let you know when it is done." }
+          hint: { text: "You can upload up to #{FileUploadValidator::MAX_FILES} files. Each file must be smaller than #{max_allowed_file_size}. You cannot upload zip files." }
       ) %>
 
       <%= f.govuk_submit "Save and continue" %>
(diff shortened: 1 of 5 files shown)


app/views/public_referrals/allegation_evidence/upload/edit.html.erb
```diff
@@ -34,7 +34,7 @@
           :evidence_uploads,
           multiple: true,
           label: { text: "Upload files", size: "m" },
-          hint: { text: "You can upload up to #{FileUploadValidator::MAX_FILES} files. Each file must be smaller than #{max_allowed_file_size}. Larger files may take longer to upload. We will let you know when it is done." }
+          hint: { text: "You can upload up to #{FileUploadValidator::MAX_FILES} files. Each file must be smaller than #{max_allowed_file_size}. You cannot upload zip files." }
```
asmega, line 37: I assume you want to be both employer and MOP journeys? therefore if this file has changes I also expect changes in `app/views/referrals/allegation_evidence/upload/edit.html.erb`

db/schema.rb
```diff
@@ -89,46 +89,46 @@
...
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.boolean "email_known"
-    t.string "email_address", limit: 256
     t.string "first_name"
     t.string "last_name"
     t.string "previous_name"
     t.string "name_has_changed"
     t.date "date_of_birth"
+    t.string "has_qts"
```
asmega, line 97: if you haven't added any migrations I don't recommend you submit schema changes as part of your PR. if you are out-of-sync you can load the schema to match column ordering

asmega:
does the following file also need changing? `app/views/referrals/allegation_previous_misconduct/detailed_account/edit.html.erb`

asmega (approved):

PR #1067, opened 2024-03-21
Fix Static pages

asmega:
# Context

- Static pages has an issue with recent rails upgrade

# Changes

- Convert static pages from markdown to use erb
- The markdown conversion process has been removed
- There seems to be an issue with processing erb with markdown from the rails upgrade hence the remove of markdown and conversion to erb as the simplest fix
- Added test coverage of the static pages to patch the hole in missing test coverage


richardpattinson (approved):
Looks great; can be shipped

PR #1089, opened 2024-04-17
Fixed build failures

VenkataChalla554:
Fixed build failures to  handle null vaules

### DIFF
--- a/terraform/statuscake.tf
+++ b/terraform/statuscake.tf
@@ -25,7 +25,6 @@ resource "statuscake_ssl_check" "domain-alert" {
 
   count = length(var.statuscake_ssl_contact_group) > 0 ? 1 : 0
 
-
   check_interval   = 3600 # Check once per hour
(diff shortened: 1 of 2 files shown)


terraform/statuscake.tf
```diff
@@ -23,7 +23,7 @@ resource "statuscake_uptime_check" "alert" {
 
 resource "statuscake_ssl_check" "domain-alert" {
 
-  count = length(var.statuscake_ssl_contact_group) > 0 ? 1 : 0
+  count = var.statuscake_ssl_contact_group != null && length(var.statuscake_ssl_contact_group) > 0 ? 1 : 0
```
saliceti, line 26: The default is [] so the value should never be null

VenkataChalla554: Yes, agreed , will remove this.

saliceti (approved):

PR #1156, opened 2024-08-09
Update Rubocop GOVUK to 5.0.2

felixclack:
We are a couple of major versions behind the GOVUK Rubocop rules.

This change updates the linter and applies the corrections these new
rules introduce.


richardpattinson (approved):
Looks good to me =)

PR #1211, opened 2024-10-17
AKS pipeline for RSM review app

neillturner:
### Context

As part of rsm migration to AKS we need to be able to build and deploy rsm to AKS as a review app automatically via a github actions pipeline.

add in .github the actions and update workflow required. 
output is made an array even though only one url to fit in with the generic scripts for smoke testing.
based on similar approach done in https://github.com/DFE-Digital/access-your-teaching-qualifications

### Guidance to review

When a change to rsm can be pushed to the rsm repo and the workflow to build and deploy the app to AKS as review app works. This should co-exist with the existing rms app (non-AKS) current workflow.
Can be tested by after deploying app is accessible https://refer-serious-misconduct-pr-1234.test.teacherservices.cloud/
also use kubectl to see that refer-serious-misconduct deployment and pods are successfully running

### Link to Trello card

[RSM: add review apps to pipeline](https://trello.com/c/YpvpP3P5/2079-rsm-add-review-apps-to-pipeline)

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/.github/actions/deploy-environment-aks/action.yml
+++ b/.github/actions/deploy-environment-aks/action.yml
@@ -1,0 +1,56 @@
+name: Deploy environment to AKS
+description: Deploys an application environment to AKS
+
+inputs:
+  environment:
+    description: The name of the environment
+    required: true
+  image-tag:
+    description: The image tag to deploy
+    required: true
+  azure-credentials:
+    description: JSON object containing a service principal that can read from Azure Key Vault
+    required: true
+  pull-request-number:
+    description: The pull request number which triggered this deploy.
+    required: false
+
+outputs:
+  environment_url:
+    description: The base URL for the deployed environment
+    value: ${{ steps.set_outputs.outputs.ENVIRONMENT_URL }}
+
+runs:
+  using: composite
+
+  steps:
+    - uses: hashicorp/setup-terraform@v3
+      with:
+        terraform_version: 1.6.4
+        terraform_wrapper: false
+
+    - uses: DFE-Digital/github-actions/set-kubelogin-environment@master
+      with:
+        azure-credentials: ${{ inputs.azure-credentials }}
+
+    - name: Terraform Apply
+      shell: bash
+      run: |
+        make ci ${{ inputs.environment }} aks-terraform-apply
+      env:
+        DOCKER_IMAGE_TAG: ${{ inputs.image-tag }}
+        PR_NUMBER: ${{ inputs.pull-request-number }}
+
+    - name: Extract Terraform outputs
+      shell: bash
+      id: set_outputs
+      run: |
+        environment_url=$(terraform -chdir=terraform/application output -raw url)
+        echo "ENVIRONMENT_URL=$environment_url" >> $GITHUB_OUTPUT
+
+    - name: Run smoke tests
+      shell: bash
+      run: |
+        environment_url=$(terraform -chdir=terraform/application output -raw url)
+        echo "Check health for $environment_url/health/all.json..."
+        curl -sS --fail "$environment_url/health/all.json" > /dev/null && echo "Health check passed for $environment_url" || echo "Health check failed for $environment_url"
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -26,7 +26,7 @@ on:
 jobs:
   build_image:
     name: Image build and push
-    if: contains(github.event.pull_request.labels.*.name, 'deploy') || github.event_name != 'pull_request'
+    if: contains(github.event.pull_request.labels.*.name, 'deploy') || contains(github.event.pull_request.labels.*.name, 'deploy-aks') || github.event_name != 'pull_request'
     runs-on: ubuntu-latest
     outputs:
       image_name_tag: ${{ steps.build_image.outputs.ghcr_image_name_tag }}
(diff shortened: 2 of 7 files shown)


neillturner:
[RSM: add review apps to pipeline](https://trello.com/c/YpvpP3P5/2079-rsm-add-review-apps-to-pipeline)

.github/workflows/build-and-deploy.yml
```diff
@@ -75,6 +75,46 @@ jobs:
...
+
+      - uses: ./.github/actions/deploy-environment-aks
+        id: deploy_aks
+        with:
+          environment: aks-review
+          image-tag: ${{ github.sha }}
+          azure-credentials: ${{ secrets.AZURE_CREDENTIALS }}
+          pull-request-number: ${{ github.event.pull_request.number }}
+
+      - uses: azure/login@v2
```
saliceti, line 98: I think azure/login and Set kubelogin environment are not required

terraform/application/output.tf
```diff
@@ -1,3 +1,5 @@
-output "url" {
-  value = module.web_application.url
+output "urls" {
```
saliceti, line 1: There is only 1 URL, so do we need the array?

neillturner: well then i would have to change the generic smoke testing job that takes an an array to take a string so its 6 or one half dozen of the other

.github/workflows/build-and-deploy.yml
neillturner: i can take them out i think they were just for the database update that we removed

saliceti: Indeed. Please remove

terraform/application/output.tf
saliceti: Yes please change on https://github.com/DFE-Digital/refer-serious-misconduct/pull/1211/files#diff-664d4eff1b79cc34f54c6f3ff518f8163e1d8e678beb35cab03a17ed1768c893R48 It will be simpler

.github/workflows/build-and-deploy.yml
neillturner: i will change it

.github/actions/deploy-environment-aks/action.yml
```diff
@@ -0,0 +1,61 @@
...
+      env:
+        DOCKER_IMAGE_TAG: ${{ inputs.image-tag }}
+        PR_NUMBER: ${{ inputs.pull-request-number }}
+
+    - name: Extract Terraform outputs
+      shell: bash
+      id: set_outputs
+      run: |
+        access_url=$(terraform -chdir=terraform/application output -json urls | jq -r '.[0]')
+        echo "ACCESS_URL=$access_url" >> $GITHUB_OUTPUT
```
saliceti, line 49: ACCESS_URL was for access-your-teaching-qualifications. Can we rename to ENVIRONMENT_URL to make it consistent with the output?

terraform/application/output.tf
neillturner: completed

.github/actions/deploy-environment-aks/action.yml
neillturner: completed

.github/workflows/build-and-deploy.yml
```diff
@@ -75,6 +75,37 @@ jobs:
...
+
+      - uses: ./.github/actions/deploy-environment-aks
+        id: deploy_aks
+        with:
+          environment: aks-review
+          image-tag: ${{ github.sha }}
+          azure-credentials: ${{ secrets.AZURE_CREDENTIALS }}
+          pull-request-number: ${{ github.event.pull_request.number }}
+
+      - name: Post comment to Pull Request $${{ env.PR_NUMBER }}
```
saliceti, line 98: The message https://github.com/DFE-Digital/refer-serious-misconduct/pull/1211#issuecomment-2419455059 doesn't show the full URL

terraform/application/config/review.tfvars.json
```diff
@@ -5,6 +5,6 @@
   "enable_postgres_ssl": false,
   "webapp_startup_command": [
     "/bin/sh", "-c",
-    "bundle exec rails db:schema_load_or_migrate && bundle exec rails server -b 0.0.0.0"
+    "bundle exec rails db:schema_load_or_migrate && bundle exec rails runner \"%i(eligibility_screener referral_form).each {|flag| FeatureFlags::FeatureFlag.activate(flag)}\""
```
saliceti, line 8: Can we add `&& bundle exec rails server -b 0.0.0.0`? or rails won't start serving requests

vipin-dfe: Yeah good catch. Fixed.

.github/workflows/delete-review-app.yml
```diff
@@ -129,3 +130,46 @@ jobs:
...
+        env:
+          PR_NUMBER: ${{ env.PR_NUMBER }}
+
+      - name: Post Pull Request Comment
+        if: ${{ github.event_name == 'pull_request' }}
+        uses: marocchino/sticky-pull-request-comment@v2
+        with:
+          header: aks
+          message: |
+            Review app  track and pay deployed to <https://refer-serious-misconduct-${{ env.PR_NUMBER }}.test.teacherservices.cloud> was deleted
```
saliceti, line 175: "track and pay" is probably from copy paste error!

neillturner: good spot correcting

neillturner:
close to delete workflow

saliceti (approved):

PR #1426, opened 2026-04-20
Replace slack notifications with teams for workflows

rgwenter-mobi:
### Trello card

https://trello.com/c/5mophRLT/2659-add-teams-channel-gh-notifications 

### Context

Remove slack notifications from workflows
Add any missing teams notifications

Workflow updates

### Guidance to review

see successful workflows run by this PR

### DIFF
--- a/.github/workflows/backup-db.yml
+++ b/.github/workflows/backup-db.yml
@@ -64,7 +64,6 @@ jobs:
             BACKUP_FILE=${BACKUP_FILE}
           fi
           echo "BACKUP_FILE=${BACKUP_FILE}" >> $GITHUB_ENV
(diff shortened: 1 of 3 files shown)


.github/workflows/backup-db.yml
```diff
@@ -72,15 +66,6 @@ jobs:
           echo "BACKUP_FILE=${BACKUP_FILE}" >> $GITHUB_ENV
           echo "KEYVAULT_NAME=${AZURE_RESOURCE_PREFIX}-${SERVICE_SHORT}-${CONFIG_SHORT}-inf-kv" >> $GITHUB_ENV
```
rgwenter-mobi: Undecided if KEYVAULT_NAME instances should be left in or removed...? 
No longer used here but does it fit in as a standard block that'll be there if wanted?

RMcVelia: Might as well remove it, it can always be added back if we need to access another secret later.

RMcVelia (approved):

